PR #9, opened 2018-09-26
Updates to configuration to enable testing

pazzarpj:
Several updates to unit tests and configuration files to enable Azure-pipelines and pytest setup


pazzarpj:
Build fails on Python 3.4 due to windows not having a package for sip. (Due to PyQt5)
Going to remove python 3.4 tests until a solution to this is made this by either removing PyQt5 dependancy by having gui as a separate package or by  removing python 3.4 as a supported version.

Issue #12, opened 2018-09-26
Modify how -c command operates and fixate.config


pazzarpj:
Currently, the -c command is used to parse in a yaml config file which is only used to set the csv path for reporting.

I am thinking we can use this as cumulative command to parse multiple config files via the command line.

This would allow arrangements where there is a default config you want set up for your environment but also want a specific set of parameters for this particular call out.

We should also consider parsing also having the ability to find a config file in the test scripts zip file.

pazzarpj:
I think this will help solve issue #11. In the context of that issue.
 
The idea is to parse in a config such as 
```yaml
part_number: 12345
plugins:
  fixate.reporting.csv:
    time_stamp_template: {0:%Y}{0:%m}{0:%d}-{0:%H}{0:%M}{0:%S}
    csv_path_template: 
      - C:\fixate_logs
      - {part_number}
      - {time_stamp_template}.csv
```
Support for setting up plug-ins with plug in specific data. 
Should give an output path of something like.
C:\fixate_logs\12345\20180927-111303.csv

clint-lawrence:
Instead of (or as well as) being able to specify multiple -c arguments, I think it would be useful to have a default config file location. Possibly %VIRTUAL_ENV%\fixate.yml.

Then when setting up a dedicated virtualenv for deployment (like we typically do) we can have some fixed global config which is tied to the installation. For more ad-hoc scenarios, the -c could be used.

[Thinking about our work deployments for a second, if there is too much tied up in the -c argument it would make it hard to run a script as it would in production without the launcher. On the other hand, having log template config decentralised isn't nice either]

I kind off like the idea of looking in cwd and/or a script zip file too, but think we should not implement that until we come up with a real use case. At the moment, we wouldn't really want a script to be able to override the log setup for example. If a that changes, happy to consider the change then.

clint-lawrence:
For a bit of inspiration, pip explains the config options nicely in it's documents (but really more flexible that we need to worry about for now)
https://pip.pypa.io/en/stable/user_guide/#configuration

pazzarpj:
Following a discussion with Clint, I think the yaml file should be flatter to make it easier to specify overriding configuration files without accidently overriding. For example what I originally proposed.
```yaml
plugins:
  fixate.reporting.csv:
    time_stamp_template: {0:%Y}{0:%m}{0:%d}-{0:%H}{0:%M}{0:%S}
    csv_path_template: 
      - C:\fixate_logs
[...]

PR #18, opened 2018-10-04
Config file updates

### LINKED ISSUE #11: fixate/reporting/csv.py Uses optional data to make directory structure
Currently, the csv file assumes that part_number and module are parsed to the program launch through --script-params.

Need to change so that the folder structure of the csv logs are able to be configured through a configuration file.

https://github.com/PyFixate/Fixate/blob/bb816960280feea263288c898ca360fbf4b7f34f/src/fixate/reporting/csv.py#L115-L122

pazzarpj:
Config file updates to resolve #11 and #12.

Added ability to have default config path as <python_env>\fixate.yml
Added -c to specify config files
Added some config helper functions for loading yaml, loading config parameters from plugins/core modules.
Changed the default csv parameters to work with default launch parameters with no config files.


src/fixate/config/helper.py
```diff
@@ -60,3 +63,85 @@ def load_json_config(in_file, config_name=None):
...
+        with open(in_file) as f:
+            if config_name:
+                fixate.config.__dict__.update({config_name: yaml.load(f)})
+            else:
+                fixate.config.__dict__.update(yaml.load(f))
+        return
+    if config_name:
+        fixate.config.__dict__.update({config_name: yaml.load(in_file)})
+    else:
+        fixate.config.__dict__.update(yaml.load(in_file))
```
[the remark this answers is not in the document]
pazzarpj: This is just importing yaml file except I kept the conventional API that The json loader and dictionary loader that already existed in that file. The extra boiler plate is so you can pass a path instead of a file pointer

clint-lawrence:
There are some other bugs lurking here. If you call a script with no config set the timestamp doesn't get templated correctly. The colons with fail on windows, but creates the file fine on os x. However we open the log in append mode. And since the file name is no longer unique, it gets added into the single log file.

screen shot 2018-10-05 at 12 33 40 am

clint-lawrence:
load_yaml_config() will fail (I think) if called with an open file handle. The os.path.exists() call should raise an exception.

pazzarpj:
> load_yaml_config() will fail (I think) if called with an open file handle. The os.path.exists() call should raise an exception.

Fixed by making it string only. if the string is valid file path, it will use the yaml library to open it, otherwise it will parse it as a string.

pazzarpj:
> There are some other bugs lurking here. If you call a script with no config set the timestamp doesn't get templated correctly. The colons with fail on windows, but creates the file fine on os x. However we open the log in append mode. And since the file name is no longer unique, it gets added into the single log file.
> 
> screen shot 2018-10-05 at 12 33 40 am

Fixed the time date stamp not rendering with default config 
Moved tpl_time_stamp and tpl_csv_path outside of plg_csv so they are independently configurable and can be used in other places.

src/fixate/config/__init__.py
```diff
@@ -29,11 +29,12 @@
 # Begin default "plugins"
 # Use plg_ prefix with dictionary of values to indicate to fixate to install this at startup
 # Default settings for csv reporting. Can be configured via yaml either removing or overriding with plg_csv
+tpl_csv_path = ["{start_date_time}-{index}.csv"]
+tpl_time_stamp = "{0:%Y}{0:%m}{0:%d}-{0:%H}{0:%M}{0:%S}"
+
```
clint-lawrence: Doesn't need fixing now, but I think we need a log file name template and a log path. I'll raise an issue and we can tidy up later.

clint-lawrence: Or maybe it's just need to be documented. See #21 for more thoughts

Issue #14, opened 2018-10-02
Make it possible to run the command line ui non-interactively


clint-lawrence:
This is initially motivated by wanting to write a few simply integration tests. In particular, I was intending to write some tests that run a simple script (that might pass, fail or error) and then check the process return code and check the generated log file.

This is complicated by the fact that the script runs assuming there will be user interaction. Of course, this makes sense in our typical deployment, but if we can decouple some of those behavioural aspects from the core logic, it will be easier to test certain parts of the code.

pazzarpj:
Agree, the final user_ok at the end of execution was to provide visual feedback to the user whether it was successful or not.
The current way we have it deployed means that the return codes created from finishing the execution are more important. 
If demand warrants the information to persist after execution, this could be added as a command line option.

PR #19, opened 2018-10-04
Removed user prompt at end of sequence

pazzarpj:
Removed user prompt at end of sequence for issue #14.

### DIFF
--- a/src/fixate/__main__.py
+++ b/src/fixate/__main__.py
@@ -282,16 +282,9 @@ class FixateWorker:
                 for task in ASYNC_TASKS:
                     task.cancel()
 
-            def finished_test_run_response(future):
-                future.result()
-                self.loop.call_soon(cancel_tasks)
-                self.loop.call_later(1, self.loop.stop)  # Max 1 second to clean up tasks before aborting
-
             def finished_test_run(future):
                 self.loop.call_soon(cancel_tasks)
-                if self.sequencer.status in ["Finished", "Aborted"]:
-                    f = partial(user_ok, "Finished testing")
-                    self.loop.run_in_executor(None, f).add_done_callback(finished_test_run_response)
+                self.loop.call_later(1, self.loop.stop)  # Max 1 second to clean up tasks before aborting
 
             init_tasks()
             self.loop.run_in_executor(None, self.sequencer.run_sequence).add_done_callback(finished_test_run)


src/fixate/__main__.py
```diff
@@ -282,16 +282,9 @@ def cancel_tasks():
                 for task in ASYNC_TASKS:
                     task.cancel()
 
-            def finished_test_run_response(future):
-                future.result()
-                self.loop.call_soon(cancel_tasks)
-                self.loop.call_later(1, self.loop.stop)  # Max 1 second to clean up tasks before aborting
-
```
clint-lawrence: I assume this was unused?

src/fixate/__main__.py
```diff
@@ -282,16 +282,9 @@ def cancel_tasks():
...
             def finished_test_run(future):
                 self.loop.call_soon(cancel_tasks)
-                if self.sequencer.status in ["Finished", "Aborted"]:
-                    f = partial(user_ok, "Finished testing")
-                    self.loop.run_in_executor(None, f).add_done_callback(finished_test_run_response)
+                self.loop.call_later(1, self.loop.stop)  # Max 1 second to clean up tasks before aborting
```
clint-lawrence: Why was this added & why only 1 second? Probably more the fact that this whole block of code needs better documenting, but I'm missing the connection between adding the self.loop.stop call where before we didn't call it at all.

pazzarpj: The async event loop chained function calls.

1. - It called finished_test_run
2. - which then scheduled user_ok to run in the event loop executor. 
3. - which had a finished callback to finished_test_run_response

The change has removed steps 2 and 3 and taken the loop stop from step 3 and put it in step 1.

The 1 second to clean up was legacy from when I tried to implement web sockets in the event loop. You need to leave time to clean up the event loop async tasks and 1 second was plenty of time to do that.

Now we have the balance of there actually being no real use of the async loop in its current form but the main loop and sequencer have been running on it for a while and I don't want to change it without having some unit tests around the functionality. I also don't know exactly how the QT GUI and all the controls Kieran put in place for cleanup of the QT GUI would be affected by changing the clean up time to the loop stopping.

src/fixate/__main__.py
pazzarpj: See other comment.
Effective functionality moved to finished_test_run

src/fixate/__main__.py
clint-lawrence: Right. Makes sense now. And agree entirely about the last paragraph

clint-lawrence:
Happy for this to be merged now

Issue #23, opened 2018-10-08
GUI doesn't Show error message in active window


pazzarpj:
The active window should show information related to the error in the active window when the error is current.

PR #26, opened 2018-10-08
Gui fixes

### LINKED ISSUE #15: GUI doesn't print user_info steps
The GUI should print user_info to the Active window until the next user blocking call.

For eg.
```python
user_info("Hello")
user_info("World")
user_ok("You got that?")
```
Currently displays
You got that?

When it should display
Hello
World
You got that?

pazzarpj:
#15 #23 
Added user_info and current error messages to Active Window


pazzarpj:
Still doesn't clear active window on change but at least presents the information.
This is a first step but doesn't solve the issues completely

clint-lawrence:
I can't spot any obvious problems. I think we need to get these fixes out there, then some tests in place so we can more confidently refactor and tidy up.

PR #38, opened 2018-10-14
Sequencer cleanup

clint-lawrence:
Hopefully this is easy to review. The intention was to change no behaviour except for adding non-interactive mode. If the equivalence isn't obvious in the final diff, I hope the incremental commits are clear.

I noted in a couple of the commit messages, we might end up reimplementing some of the things that I think the code I removed was originally trying to achieve. Until that time, I'm keen to remove complexity where possible.


clint-lawrence:
One other thing. Once this is merged, if we need to go back to blocking at the end of test scripts instead of existing immediately, we can do so based on the --non-interactive argument.

pazzarpj:
All looks good. All just legacy code that is no longer used.

pazzarpj:
Tests pass and manual checks seem to have the same behaviour

PR #39, opened 2018-10-17
Basic integration tests

clint-lawrence:
With the new non-interactive mode command line parameter, we can now get some more useful integration tests working. This first PR adds two simple tests and a new command line parameter to set the log file output.


pazzarpj:
All good

PR #42, opened 2018-10-22
Add additional script to log integration test

clint-lawrence:


clint-lawrence:
@pazzarpj, I just remembered the CI doesn't run of forked repos. Should I push the same branch to the main repo?

pazzarpj:
@clint-lawrence I fixed it so it does work on forks. The bug is still there but I don't publish the wheel.

pazzarpj:
Failing test_basichierachy[None-test_test-None-10] on python 3.7 only
At index 8 diff: ['Test 1.1', 'exception', "AssertionError('Exception in the test test',)"] != ['Test 1.1', 'exception', "AssertionError('Exception in the test test')"]

pazzarpj:
https://docs.python.org/3/whatsnew/3.7.html

repr for BaseException has changed to not include the trailing comma. Most exceptions are affected by this change. (Contributed by Serhiy Storchaka in bpo-30399.)

pazzarpj:
Fixed that. Can you review if that's the way you want the issue resolved?

pazzarpj:
After that, happy to merge

clint-lawrence:
It exposes the fragility of these tests a little. Although there shouldn't be too many places where we're at the mercy of external behaviour in the log, so hopefully similar things to bite us much.

Given that the repr isn't that useful a better solution long term might be to not use the repr, just define the message we want to log and the create a separate error log which a full traceback. But other than those random thoughts, your fix is fine.

Issue #70, opened 2019-01-28
Jig Driver isn't correctly throwing an error


pazzarpj:
Error should be thrown if pin is reused in multiple multiplexers.
RKS Driver showed that this isn't the case

pazzarpj:
The error should be raised as a multiplexer should have exclusive access to the pin.
Where multiple multiplexing operations need to happen at the same time. Ensure that trigger_update=False is parsed to the muxing functions.

This should be nicer to use if we create mux groups which will send multiple signals at once to a group of multiplexers to emulate the behavior without duplicating the pin names

PR #71, opened 2019-01-29
Fix for pin duplicates in jig driver

pazzarpj:
#70 Duplicate pin values should raise an appropriate error noting where the pin is used multiple times.

### DIFF
--- a/src/fixate/core/jig_mapping.py
+++ b/src/fixate/core/jig_mapping.py
@@ -20,6 +20,10 @@ class VirtualAddressMap:
         return list(zip(self.virtual_pin_list,
                         bits(self._virtual_pin_values, num_bits=len(self.virtual_pin_list), order="LSB")))
 
+    def active_pins(self):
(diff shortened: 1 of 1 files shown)


pazzarpj:
Code to replicate issue
```python
from fixate.core.jig_mapping import JigDriver, AddressHandler, VirtualMux


class Mux1(VirtualMux):
    pin_list = ["K1", "K2"]

class Mux2(VirtualMux):
    pin_list = ["K3", "K4"]

class Mux3(VirtualMux):
    pin_list = ["K2", "K5"]

class Addr(AddressHandler):
    pin_list = ["K{}".format(x) for x in range(1, 10)]

class Jig(JigDriver):
    address_handlers = [Addr()]
    multiplexers = [Mux1(), Mux2(), Mux3()]

jig = Jig()
```

pazzarpj:
Fixes for #70 and #72

Issue #56, opened 2018-12-04
Fix gui imports so fixate runs on a headless Rpi


clint-lawrence:
fixate should be able to run without the gui. That includes not importing gui libraries and having dependencies on X. When trying to run on a headless Rpi fixate was failing without having an X server running.

clint-lawrence:
I think the problem is in the pynput module. This is used to capture keyboard events.
```
File "/home/pi/fixate-venv/lib/python3.5/site-packages/fixate/ui_cmdline/cmd_line.py", line 7, in <module>
    from pynput import keyboard
File "/home/pi/fixate-venv/lib/python3.5/site-packages/pynput/__init__.py", line 23, in <module>
```

pazzarpj:
Cool. We can move it to requirements_test.txt

pazzarpj:
Actually, I think that was done by Kieran for the keyboard hooks. Might just need to find another cross platform solution.

clint-lawrence:
Yep, exactly. Thinking it might just be easiest to switch back to termios for os x/linux and msvcrt on windows.

clint-lawrence:
This might work. Uses termios/msvcrt.
https://github.com/magmax/python-readchar

smontanari1:
Readchar is suitable alternative to pynput or msvcrt but does not have a function for testing if a key is being pressed, only recording the next keystroke and will block until you press a key. Need help to examine if possible to implement

PR #76, opened 2019-01-31
Keyboard fixes

smontanari1:
Fixes dependency on msvcrt by using a script which chooses between msvcrt and termios implementation of keyboard hit, tested and confirmed working on headless raspberry pi over ssh

#56
#1

### DIFF
--- a/src/fixate/ui_cmdline/cmd_line.py
+++ b/src/fixate/ui_cmdline/cmd_line.py
@@ -4,7 +4,7 @@ import time
 import textwrap
 from queue import Empty
 from pubsub import pub
-from .kbhit import KBHit
+from fixate.ui_cmdline.kbhit import KBHit
 from queue import Queue
 from fixate.core.exceptions import UserInputError
 from fixate.core.common import ExcThread
--- a/src/fixate/ui_cmdline/cmd_line.py
+++ b/src/fixate/ui_cmdline/cmd_line.py
@@ -4,6 +4,7 @@ import time
 import textwrap
 from queue import Empty
 from pubsub import pub
+from .kbhit import KBHit
 from queue import Queue
 from fixate.core.exceptions import UserInputError
 from fixate.core.common import ExcThread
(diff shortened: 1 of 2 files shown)


src/fixate/ui_cmdline/cmd_line.py
```diff
@@ -1,67 +1,61 @@
...
+import time
 import textwrap
-import traceback
 from queue import Empty
-from queue import Queue
 from pubsub import pub
-from pynput import keyboard
-import fixate.config
-from fixate.config import RESOURCES
+from .kbhit import KBHit
```
pazzarpj: I think this should be an absolute import
```python
from fixate.ui_cmdline.kbhit import KBHit
```

src/fixate/ui_cmdline/kbhit.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
+'''
+A Python class implementing KBHIT, the standard keyboard-interrupt poller.
```
pazzarpj: Include where you found this code snippet into this doc string.

src/fixate/ui_cmdline/cmd_line.py
```diff
@@ -1,67 +1,61 @@
...
 from fixate.core.exceptions import UserInputError
+from fixate.core.common import ExcThread
+from fixate.config import RESOURCES
+import fixate.config
 
 cmd_line_queue = Queue()
 wrapper = textwrap.TextWrapper(width=75)
 wrapper.break_long_words = False
 
 wrapper.drop_whitespace = True
```
pazzarpj: This is all just a revert of the changes made during #1.
Approved because it should be identical functionality

smontanari1: Do you want these lines included? they are removed in another change you made

pazzarpj: Keep them removed. The comment was meant to be from the full revert of the changes that Kieran made in #1.

Issue #51, opened 2018-11-19
FTDI driver doesn't work on 64 bit


pazzarpj:
```python
from fixate.drivers import ftdi
f = ftdi.open
```
Gives a cannot set baud rate error.
This works on 32 bit python.

pazzarpj:
This can also cause python to crash

pazzarpj:
I suspect the original issue had nothing to do with 64 vs 32 bit but a corrupted dll files on the pc.
Further investigation into the original case that raised this issue that several other events could cause the gui to crash on the same computer that couldn't be replicated on multiple other computers.

pazzarpj:
New information shows that it works on 64 bit Windows 7 and Linux but not on Windows 10 64 bit

PR #80, opened 2019-02-05
Ftdi fixes

pazzarpj:
Fixes for #51

### DIFF
--- a/src/fixate/drivers/ftdi.py
+++ b/src/fixate/drivers/ftdi.py
@@ -53,13 +53,18 @@ class FT_STATUS(object):
     FT_NOT_SUPPORTED = DWORD(17)
     FT_OTHER_ERROR = DWORD(18)
 
-    def get_status_message(self, value):
-        for msg, itm in FT_STATUS.__dict__.items():
-            try:
-                if itm.value == value:
-                    return msg
-            except AttributeError:
-                pass
+
+def get_status_message(value):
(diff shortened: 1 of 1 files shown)


pazzarpj:
@smontanari1 Can you please confirm that these fixes work on windows 7 64 bit as well?

src/fixate/drivers/ftdi.py
```diff
@@ -54,6 +57,19 @@ class FT_STATUS(object):
...
+    """
+    :param value: Integer representing the error message
+    :return:
+    """
+    for msg, itm in FT_STATUS.__dict__.items():
+        try:
+            if itm.value == value:
+                return msg
+        except AttributeError:
+            pass
```
clint-lawrence: This either needs a comment to justify why it's O.K. to ignore the exception, or improved in some other way.

Assuming that msg is a string, returning a empty string or something like f"No error message found for error number '{value}'"?

pazzarpj: Changed FT_STATUS to a dictionary as it isn't used outside of error checking. (ie. the DWORD ctypes are never needed or used.)
```python
class FTD2XXError(BaseException):
    pass


def check_return(return_code):
    if return_code != 0:
        raise FTD2XXError(FT_STATUS.get(return_code, "FT_UNKNOWN_ERROR"))
```

Issue #58, opened 2018-12-19
Post Sequence info is out of order


pazzarpj:
The post sequence info can be out of order from the original calling if you have a mix of user_post_sequence_info, user_post_sequence_info_fail, user_post_sequence_info_fail.

This is due to an individual list maintained for each type of post sequence call.

Suggest unifying.

Issue #43, opened 2018-10-25
Post Sequence Info not viewable in Active Window


pazzarpj:
Post sequence info is not present in the active window at the end of testing.

Obviously this should be present

PR #84, opened 2019-02-17
Gui refactor

pazzarpj:


pazzarpj:
Basic fixes for #58 #43.

clint-lawrence:
Looks O.K. to me.
One thing that occurred to me (to consider later). Instead of having user_post_sequence_* functions, there could be a post_sequence hook function. (and pre-sequence etc) That could have useful sequence data like the number of pass/fail tests passed in so info to print could be customised at the time of writing to the user.

Issue #65, opened 2019-01-25
DSO driver has many modes of failure


pazzarpj:
Add fixes for measurements causing timeout errors when data isn't available

pazzarpj:
Closing this until a specific example arises after the current round of fixes.

PR #85, opened 2019-02-18
Discovery fixes and scope fixes

### LINKED ISSUE #61: Auto detect over serial causes issues
The auto detect writes data to the available com ports.
This can cause issues, especially when these com ports are connected to FTDI chips that are bit banging the resultant data.

This has caused relays to toggle and break the system.

Disable all the auto detect related to serial ports and work out an alternate way of reliably specifying connected instruments without accidentally causing problems.

pazzarpj:
#61
Removed auto detect in general. Removed manipulation of config.

#65 
Many improvements related to DSO single acquisition with automatic waits for arming of trigger, polling for trigger and checking complete acquisition.
Single mode blocks until trigger is armed.

### DIFF
--- a/src/fixate/__main__.py
+++ b/src/fixate/__main__.py
@@ -10,7 +10,6 @@ from zipimport import zipimporter
 from pubsub import pub
 import fixate.config
 from fixate.config import ASYNC_TASKS, RESOURCES
-from fixate.config.local_config import save_local_config
 from fixate.core.exceptions import SequenceAbort
 from fixate.core.ui import user_input, user_serial, user_ok
 from fixate.reporting import register_csv, unregister_csv
@@ -283,7 +282,6 @@ class FixateWorker:
             unregister_csv()
             if serial_number == "ABORT_FORCE" or test_selector == "ABORT_FORCE":
                 return 11
-            save_local_config()
             self.clean = True  # Let the supervisor know that the program is finishing normally
             if self.sequencer.end_status == "FAILED":
                 return 10
--- a/src/fixate/core/discover.py
+++ b/src/fixate/core/discover.py
@@ -169,7 +169,7 @@ def _visa_id_query(instrument):
         if resp:
             instr.close()
             del instr
-            return resp, instrument
+            return [resp, instrument]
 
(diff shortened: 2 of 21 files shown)


src/fixate/__main__.py
```diff
@@ -10,7 +10,6 @@
 from pubsub import pub
 import fixate.config
 from fixate.config import ASYNC_TASKS, RESOURCES
-from fixate.config.local_config import save_local_config
```
clint-lawrence: Can you delete save_local_config also?

src/fixate/core/discover.py
```diff
@@ -57,27 +54,13 @@ def discover_sub_classes(sub_class_match):
...
     except Exception as e:
-        pass
-        # print(e)
-        # print("Import Error: {}".format(module))
-    else:
-        mod = sys.modules[module]
-        return inspect.getmembers(mod, inspect.isclass)
-    return []
-
+        return []
```
clint-lawrence: Doesn't look like this (and discover_classes above) are used anymore. Can we delete?

src/fixate/drivers/dso/helper.py
```diff
@@ -12,7 +12,7 @@
     number = float
 
 
-def open(restrictions=None):
+def open():
```
clint-lawrence: All the driver/*/helper.py files reference discover_sub_classes. I'm not sure if the discover functions get used anywhere, but I think discover_sub_classes is broken.

discover_sub_classes references fixate.config.CLASS_LIST. As far as I can tell, it only ever gets initialised to None. I'm 100% though, since there might be some non-obvious reference. But possibly another item that can be cleaned up?

src/fixate/core/discover.py
pazzarpj: Yeah, that was used before i had DRIVER_LIST

src/fixate/drivers/dso/helper.py
pazzarpj: Yeah it can be cleaned up. DRIVER_LIST was added to explicitly define the implemented classes which made the discover classes unused.

Issue #83, opened 2019-02-14
DMM Diode mode is broken


pazzarpj:
DMM diode mode sets
```python
 def _build_mode_string(self):
        """
        Creates list of configuration strings and combines into a single string
        return: Configuration string
        """
        build_list = [self._modes.get(self._mode, ""), self._range_string, self._rtd_type,
                      self._diode_str]
        build_string = " ".join([x for x in build_list if x != ''])

        return build_string
``` 
Once diode_str is set, all future commands break as it isn't cleared. Same with rtd_type.

Diode mode also needs a comma between the two commands.

```python
def diode(self, low_current=True, high_voltage=False):
        """
        Writes configuration string for diode to the DMM
        param _range: value set for the range
        param _resolution: value set for the resolution
        """
        self._diode_str = "{} {}".format(int(bool(low_current)), int(bool(high_voltage)))
        self._set_measurement_mode('diode')
```
needs to be
```python
        self._diode_str = "{}, {}".format(int(bool(low_current)), int(bool(high_voltage)))
```

PR #88, opened 2019-02-22
Dmm diode

pazzarpj:


pazzarpj:
#83 Plus some refactor and removing of code.
Removing :SYS:ERR? calls after each measurement as it isn't needed and slows down acquisition.

src/fixate/drivers/dmm/helper.py
```diff
@@ -14,35 +13,22 @@ def open():
...
-    def __init__(self, instrument):
-        self.instrument = instrument
-        self.samples = 1
-
-    @abstractmethod
-    def measure(self, *mode, trigger=True, **mode_params):
-        pass
-
-    @abstractmethod
     def measurement(self):
```
clint-lawrence: Raise NotImplementedError instead of pass, now that we're not using @abstractmethod

PR #91, opened 2019-03-22
Update in response to dmm-diode pull request feedback

noplacebo:
Add @wraps and import from functools to deprecated in core/common.py
unittest added in test/core/test_wraps.py
Request @clint-lawrence review changes


clint-lawrence:
Thanks for adding the test. As we discussed, the test can be improved by focusing on the big picture of what the decorator is doing, rather than getting too focused on this specific change. As with all tests, try to focus on the things that are fundamental to the functionality, rather than implementation. So:
- Does the decorator return a callable?
- When we call the return value of the decorator, does our original function get called?
- Does the warnings.warn function get called?

Also, rename the test file to test_deprecated.py, since that is what we are actually testing.

Have a look in the documentation for the warnings module. There is a context manager that allows easy testing that warnings get called:
https://docs.python.org/2/library/warnings.html#testing-warnings

One more thing. Some of the tests use the unittest module, but we are actually using pytest to run tests. For new tests it is preferable to use pytest instead. see test_script2log.py for examples.

PR #92, opened 2019-03-25
Experiment with using tox to run test on CI & local.

clint-lawrence:
The pytest invocation details are moved from azure-pipelines.yml to
tox.ini. tox.ini now uses requirements-test to install pytest &
required plugins.

Creating a pull request to kick off a build...


clint-lawrence:
I think it all works... You want to have a look over and see if it looks O.K.? I'll also run tox locally on windows when I get to work in the morning, to check the test run locally without any issue.

PR #106, opened 2019-05-08
Enable Black style check on CI

clint-lawrence:
I'll confirm that the check fails on CI and then follow up with a commit that includes the reformatting


clint-lawrence:
@pazzarpj & @noplacebo if there are no objections I'll merge and then get ready to release 0.5.1 with only these formatting changes some time tomorrow (hopefully...)

Issue #50, opened 2018-11-14
User action doesn't allow you to pass on GUI


pazzarpj:
User action on cmd_line says space for pass and escape for fail.
GUI needs a pass fail button for the user_action

clint-lawrence:
With #108, the cmd_line actually changed, so space no longer forces a pass. (and the prompt was change accordingly).

I don't think this is a problem anywhere it has been used to date, and seemed to be a more consistent with the concept of the user_action call. So, closing this now essentially as a won't fix. Happy to re-introduce that capability in the future if a need arises.

Issue #54, opened 2018-11-26
User action causes the next user command to hang GUI


pazzarpj:
example

```python
def my_test():
    return True

user_ok("This displays")
user_action("Display This", my_test)
user_ok("This hangs the GUI")
```

pazzarpj:
![image](https://user-images.githubusercontent.com/6457830/49118949-84e2c500-f2fa-11e8-82b4-92cadd8dbac6.png)

Log from Autocard test jig ran in GUI.

Seems to cause a similar issue at the end of the sequence which prompts test finished. Will cause the exit code to show error even though the test was successful.

clint-lawrence:
I can't replicate on master. The full test script I'm using is: 
```
"""
This is a test script that shows basic use case for the fixate library
"""
from fixate.core.common import TestClass
from fixate.core.ui import user_action, user_ok

__version__ = '0'


class ATest(TestClass):
    """
    Asks the user to push the red button
    """
    test_desc = "A Test"

    def test(self):
        class MyTest:
            def __init__(self):
                self.count = 0

            def my_test(self):
                user_ok(str(self.count))
                self.count += 1
                return self.count > 3

        c = MyTest()
        user_ok("ok: This displays")
        user_action("action: Display This", c.my_test)
        user_action("action: Display This", lambda: True)
        user_ok("ok: This hangs the GUI")


TEST_SEQUENCE = [ATest(), ATest()]
```

@pazzarpj, any thoughts on how to replicate?

pazzarpj:
My original code block I think used to cause the issue. I'm a bit out of action to check others at the moment

clint-lawrence:
@pazzarpj will add more info soon, but it's looking like a dependency issue

clint-lawrence:
When I couldn't replicate the failure above, there was a bug in my code. I'd started clean up some things in user_action. But in doing so, I made a type and was sending message topic "UI_**A**ction", instead of "UI_**a**ction". The correct topic is with a lower case "a".

So after fixing that issue, I can replicate the bug as described. In thinking about the problem, I've
also come to the conclusion that there are two fairly distinct usages of user_action. We may or may not be able to accommodate both when fixing the bug. The target function supplied to user_action can do two different things:
1.  Interact with the user with other function call. e.g. calling user_ok or user_yes_no. I'll have to review some places where this is used, but I'm not sure we even need a user_action in this case. We could just use a loop in the test script.
[...]

PR #108, opened 2019-05-16
Ui action fix

clint-lawrence:
Still need to fix the bug... But have spend some time tidying the code.

### DIFF
--- a/src/fixate/core/ui.py
+++ b/src/fixate/core/ui.py
@@ -38,24 +38,6 @@ def _user_req(msg):
     return resp
 
 
-def _user_image(path):
-    """
-    A GUI function that updates the displayed image on-screen
-    :param path:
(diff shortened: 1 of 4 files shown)


clint-lawrence:
@pazzarpj getting some eyes on this would be really good.
Two things I'm not sure about:
1. Should we be sending UI_block_start/end around user_action? I think yes but we didn't before either.
2. If we call user_yes_no or user_ok functions in the target of a user_action, they will clear the "Fail" button. I think in reality that is getting to the point where the whole intent of user_action is getting broken down, so might no be a big concern.

clint-lawrence:
@pazzarpj pending any feedback you have, I believe this now fixes #50 and #54. Would appreciate a second set of eyes before merging.

PR #120, opened 2020-08-06
fix formatting to with black

clint-lawrence:


clint-lawrence:
Merging despite the build failure. Will follow up with a new branch for that fix

PR #122, opened 2020-08-07
Fixed incompatible signatures of poutput for new versions of cmd2 

### LINKED ISSUE #119: Version 0.9.14 of cmd2 has breaks fxconfig in version 0.5.3 of fixate
https://github.com/python-cmd2/cmd2/releases/tag/0.9.14

perror & poutput no longer accept color keyword arguments, which are used in fxconfig.

to work around, manually install 0.9.13 instead. i.e.
```pip install cmd2==0.9.13```

### PR DESCRIPTION
**Pull request by daniel-montanari**
Replaced manual ansi escape codes with cmd2.ansi module

Resolves PyFixate/Fixate#119

### DIFF
--- a/src/fixate/core/config_util.py
+++ b/src/fixate/core/config_util.py
@@ -5,7 +5,7 @@ import json
 import copy
 from shutil import copy2
 from pathlib import Path
-from cmd2.ansi import Fore
+from cmd2.ansi import style, fg
 from fixate.drivers.pps.bk_178x import BK178X
 import fixate.config
 from pyvisa.errors import VisaIOError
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/fixate/core/config_util.py [hunk 1]
```diff
@@ -5,6 +5,7 @@
 import copy
 from shutil import copy2
 from pathlib import Path
+from cmd2.ansi import Fore
```
clint-lawrence, line 8: I don't think we should be import Fore like this. It's not part of the publish api docs. We could use it if we import direct from colorama

src/fixate/core/config_util.py
```diff
@@ -305,14 +300,14 @@ def _print_config_dict(self, config_dict):
             self.poutput("SERIAL || " + com_port + " || " + str(parameters))
 
     def _test_print_error(self, name, msg):
-        self.poutput("ERROR: ", end="", color=RED)
-        self.poutput(str(name), end="", color=CYAN)
-        self.poutput(" - {}".format(msg))
+        self.poutput(Fore.RED + "ERROR: ", end="")
+        self.poutput(Fore.CYAN + str(name), end="")
+        self.poutput(Fore.WHITE + " - {}".format(msg))
```
clint-lawrence, line 305: instead of directly concatenating the style, I think we should use the api provided
self.poutput(style("ERROR: ", fg="red"), end="") etc

That docs aren't super clear on what is expected...

clint-lawrence (commented):
Can you also test this works with 0.9.13 of cmd2

src/fixate/core/config_util.py
daniel-montanari: Ansi module was added in 0.9.14, I've replaced Fore with fg, which is how cmd2.ansi uses Fore

src/fixate/core/config_util.py
daniel-montanari: fg argument is the result of fg_lookup, which expects either a str or fg Enum (which is essentially Fore from colorama)

PR #123, opened 2020-08-11
discover.py - simplified control flow, added pubsub message for broad…

daniel-montanari:
Basic idea implemented with the dmm drivers.
On loading a driver log the .serial property to the csv.
I haven't given too much thought yet to whether this should be protected, or whether it should just be a method instead. I think a property makes sense though as the serial number shouldn't change once the instrument has been instantiated.

Remaining drivers still need a .serial property added to them so open_visa_instrument in discover.py can use them.


src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -62,6 +62,7 @@ def __init__(self, instrument, *args, **kwargs):
         self._mode = None
         # Set to True to have explicit error checks on each read
         self.legacy_mode = False
+        self.serial = self.get_serial()
```
clint-lawrence, line 65: I would leave this out. We're not likely to call get_serial often, so if we do need it more than once it's not too big a cost to call out to the instrument multiple times.

src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -341,3 +342,11 @@ def analog_filter(self, bandwidth=None):
         elif any(x in self._mode for x in ["dc", "res"]):
             self._write(self._filters[self.mode] + ":FILT:STAT ON")
         pass
+
+    def get_serial(self):
```
clint-lawrence, line 346: I think we should either parse out the serial number or rename this to get_identity.
Possibly, rename this to get identity, the implement get_serial using this function.

src/fixate/reporting/csv.py
```diff
@@ -367,6 +377,7 @@ def register_csv():
     pub.subscribe(writer.reporting.sequence_complete, "Sequence_Complete")
     pub.subscribe(writer.reporting.user_wait_start, "UI_block_start")
     pub.subscribe(writer.reporting.user_wait_end, "UI_block_end")
+    pub.subscribe(writer.reporting.discover_visa, "discover_visa")
```
clint-lawrence, line 380: Rename this message to reflect its function rather than where we're calling it from. "driver_open" maybe?

src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -341,3 +342,11 @@ def analog_filter(self, bandwidth=None):
...
+        """
+        serial string returned by *IDN?
+        outg_sw_date is the
+        :return: FLUKE,8846A,S/N
+        """
+        return self.instrument.query("*IDN?")
```
clint-lawrence, line 352: Does query return bytes or string? I'm inclined to think we should be explicit here that get_serial (or get_identity if we implement that) should return a string and not a bytes object, otherwise we potentially need to know about the encoding used elsewhere in the code base.

src/fixate/drivers/dmm/helper.py
```diff
@@ -67,3 +67,6 @@ def diode(self, low_current=True, high_voltage=False):
 
     def reset(self):
         raise NotImplementedError
+
+    def get_serial(self):
```
clint-lawrence, line 71: I'm torn on if this should be NotImplementedError or return "". We might not be able to get a serial number from all drivers and for some, it might not make sense. I suppose the argument in favor of keeping it NotImplementedError would be to force the driver implementation to be explicit. It can alwa

clint-lawrence (commented):
To get some test coverage of this, we could use the existing script2log tests. The sticking point is finding a sensible way to inject a test driver. Having a pytest fixture for that might be useful for other purposes also...

src/fixate/drivers/dmm/helper.py
daniel-montanari: Some of the other drivers simply pass rather than raising an error.

src/fixate/drivers/dmm/fluke_8846a.py
daniel-montanari: Returns a string

src/fixate/reporting/csv.py
daniel-montanari: I'll rename it to "driver_open" and change the "VISA" string in the csv writer to "DRIVER"

src/fixate/drivers/dmm/fluke_8846a.py
daniel-montanari: I originally was doing some string parsing to get the serial number out, but realised that including everything meant that we could keep the other parameters such as model number, while still letting us filter through them. 
I'll rename it to get_identity

src/fixate/drivers/pps/bk_178x.py
```diff
@@ -355,21 +358,6 @@ def identify(self, as_string=False):
...
-    mypps.output_ch1 = True
-    mypps.output_ch1 = False
-    mypps.voltage = 24
-    mypps.output_ch1 = True
-    mypps.output_ch1 = False
-
-    mypps.output_ch1 = False
+    def get_identity(self) -> str:
+        # TODO: this won't be formatted how we want
+        return self.identify(as_string=True)
```
clint-lawrence, line 363: Note sure I follow, but I assume you mean to come back to this and fix the formatting?

src/fixate/drivers/pps/helper.py
```diff
@@ -206,3 +198,14 @@ def idn(self):
                 inspect.currentframe().f_code.co_name
             )
         )
+
+    def get_identity(self):
```
clint-lawrence, line 202: Implement this by calling idn? or the bind the name as an alias?
self.get_identify = self.idn should work...?

src/fixate/drivers/pps/bk_178x.py
daniel-montanari: The self.identify function doesn't return a comma delimited string and includes the keys. Will make it consistent later.

src/fixate/drivers/pps/helper.py
daniel-montanari: PPS was a bit of a mess, bk 178x doesn't seem to even us idn, spd 3303x only mentions it in the api list, which gives me conniptions and I was trying to avoid looking at it. Not how binding names works with inheritance either.

PR #124, opened 2020-08-18
Daqmx

### LINKED ISSUE #116: Changes to ExcThread need review in daqmx driver
The daqmx driver's TwoEdgeSeparation possibly relied upon some of the nastier aspects of the old version of ExcThread (present in 0.5.2). This most likely will need to be reviewed for robustness under error. However we will need to setup some hardware to sufficiently test.

daniel-montanari:
Reworked the daq TwoEdgeSeparation class to be compatible with ExcThread
Resolves #116

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ pyqt5
 pynput
 ruamel.yaml
 pyserial
-cmd2
+cmd2>0.9.13
 PyDAQmx==1.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ pynput
 ruamel.yaml
 pyserial
 cmd2
+PyDAQmx==1.3
--- a/src/fixate/drivers/daq/daqmx.py
+++ b/src/fixate/drivers/daq/daqmx.py
@@ -14,7 +14,6 @@ from PyDAQmx import (
     float64,
     uInt64,
     c_char_p,
-    c_char,
     uInt32,
 )
 
@@ -329,7 +328,8 @@ class TwoEdgeSeparation(DaqTask):
         self.source_terminal = source_terminal
         self.destination_terminal = destination_terminal
         self.validate_terminals = validate_terminals
-        self.error_queue = Queue()
+        self._error_queue = Queue()
+        self._thread_timeout = 10
 
     def init(self):
         if self.task_state == "":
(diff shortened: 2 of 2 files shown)


daniel-montanari:
Tested with both PyDAQmx 1.3 and 1.4.3
1.3 All works normally
1.4.3 Has a different signature for DAQmxReadCounterScalarF64 and raises an exception correctly

src/fixate/drivers/daq/daqmx.py
```diff
@@ -392,31 +396,44 @@ def init(self):
...
+            DAQmxReadCounterScalarF64(self.task, float64(10), byref(self._data), None)
+        except Exception as e:
+            self._error_queue.put(ThreadError(e))
+        return
 
     def trigger(self):
         if self._trigger_thread:
             self.clear()
-            self._trigger_thread.join(10)
+            self._trigger_thread.join(self._thread_timeout)
```
clint-lawrence, line 422: I really wish NI had better documentation... It not at all clear how to correctly use the Task api. It feels to me like the clear() and join() calls here are unnecessary. I think the clear(), which call DAQmxClearTask() actually deallocates the task. So this method should never work if the self._trigger_thread block executes...

But I am reluctant to dive too far into this right now. My gut feel is that a nice way to use the daq tasks would be to implement them as a context manager, so that clean up is done correctly.

clint-lawrence (changes requested):
Aside from the couple of little nit picks, I think we should merge this as is. I'm not convinced the way we're using the DAQ api is strictly correct, but we know the general approach has been reliable in production :(

requirements.txt
```diff
@@ -4,4 +4,5 @@ pyqt5
 pynput
 ruamel.yaml
 pyserial
-cmd2
\ No newline at end of file
+cmd2
```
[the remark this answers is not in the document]
daniel-montanari: Do you want me to add that to this pr

src/fixate/drivers/daq/daqmx.py
daniel-montanari: Clearing the task does deallocate the task, and sets the state to "", which means that when the trigger thread is created the init method recreates the task again.
I'm not sure whether the init should be done in the thread itself or whether it should be done in the trigger call.
I think that reading the result is really the only thing that needs to be in the thread. I can test it to be sure if needed.

PR #136, opened 2022-01-10
Update the scope idn regex

clint-lawrence:
Some newer scope have a slightly different
format for the model number string return by
the IDN? query.


clint-lawrence:
I've tested the driver opens with DSOX1102G, DSOX1202G and MSOX3014

PR #142, opened 2022-06-28
Fixed pyvisa imports

John2202W:
I've fixed the remaining pyvisa imports and added basic tests to import each driver


test/drivers/test_driver_imports.py
```diff
@@ -0,0 +1,51 @@
+# import unittest.mock
```
clint-lawrence, line 1: Delete all the commented out code

test/drivers/test_driver_imports.py
```diff
@@ -0,0 +1,51 @@
...
+# import sys
+# import os
+# import ctypes
+import pytest
+
+
```
clint-lawrence, line 7: I think it would be worth added a general comment that explains the basic idea here.
```
"""These tests are a very minimum smoke test on the drivers.

while we work to get a more complete test in place that includes the
hardware, these are a minimum check that the module can at least
be imported"""
```

test/drivers/test_driver_imports.py
```diff
@@ -0,0 +1,51 @@
...
+def test_driver_files():
+    import fixate.drivers.dmm
+    import fixate.drivers.dso
+    import fixate.drivers.funcgen
+    import fixate.drivers.lcr
+    import fixate.drivers.pps
+
+
+# _files_path = os.path.join(os.path.dirname(__file__), "files")
```
clint-lawrence, line 16: If we're not going to use this, I think we should delete the commented out code and remove the .h/.dll files.

Could keep this on a separate branch if you want to keep it in the repo. Hopefully Jasper's work will replace the need for all this soon enough in any case.

clint-lawrence (changes requested):

test/drivers/test_driver_imports.py
John2202W: Done

test/drivers/test_driver_imports.py
John2202W: Done

test/drivers/test_driver_imports.py
John2202W: Done

PR #144, opened 2022-07-04
Add workflow action to publish to testpypi

John2202W:
Added github action to build and publish to test pypi either upon new tag or manually.
Steps to publish to pypi are added but commented out and un-tested due to permissions.


clint-lawrence:
Hey John, shouldn’t need any other permissions. The token for pypi should be ready to go. As long as someone can merge this to master, it should be good to go.

Without a laptop, so can’t easily check more than that, but let me know if you need any permissions tweaked.

John2202W:
Yeah needs someone to merge to double-check the pypi dist. process - can't get a token on my fork to test with.

Neither Dan or I have permissions to create a release. (@christopherwallis @jcollins1983 ?)

PR #147, opened 2022-07-08
Add gui gif support

John2202W:
Added support to play .gifs in the gui.

NOTE: doesn't play nice with overlaying multiple images atm.

### DIFF
--- a/src/fixate/ui_gui_qt/ui_gui_qt.py
+++ b/src/fixate/ui_gui_qt/ui_gui_qt.py
@@ -275,7 +275,7 @@ class FixateGUI(QtWidgets.QMainWindow, layout.Ui_FixateUI):
             else:
                 # NOTE: full overlay features not available with .gif
                 #   due to forcing the scene dimensions to the movie frame
-                if len(self.image_scene.items())>0:
+                if len(self.image_scene.items()) > 0:
                     logger.error("Unsupported behaviour when overlaying .gifs")
                 animation = QtWidgets.QLabel()
                 movie = QtGui.QMovie(path)
@@ -299,7 +299,7 @@ class FixateGUI(QtWidgets.QMainWindow, layout.Ui_FixateUI):
         self.sig_image_clear.emit()
 
     def on_image_clear(self):
-        """ Create a fresh graphics scene """
+        """Create a fresh graphics scene"""
         # NOTE image_scene.clear() does not remove all properties as desired
         self.image_scene = QtWidgets.QGraphicsScene()
         self.ImageView.set_scene(self.image_scene)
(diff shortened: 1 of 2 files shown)


src/fixate/ui_gui_qt/ui_gui_qt.py
```diff
@@ -266,9 +266,27 @@ def on_image_update(self, path):
             # message dialog so the user knows the image didn't load
             self.file_not_found(path)
         else:
-            image = QtGui.QPixmap()
-            image.loadFromData(image_data)
-            self.image_scene.addPixmap(image)
+            _, ext = os.path.splitext(path)
+            # Default to normal image if not a gif
+            if ext != ".gif":
+                image = QtGui.QPixmap()
```
clint-lawrence, line 272: `ext.lower() != ".gif"` is probably better.

Also wondering if adding a new `user_ ` function would be a good idea? `user_gif` or `user_animated_gif`? Then there is no need to guess at what to do.

clint-lawrence (commented):
I know this has been on Ben's wish list for a while. Will be a good addition.

src/fixate/ui_gui_qt/ui_gui_qt.py
John2202W: Added the case-insensitive matching.
Can easily make it a separate user function - do we want it a part of the "overlaying" graphics scene? Don't know if that is ever used?

daniel-montanari: I like the idea of an animated gif function, since regular gifs should still be able to work as normal images.
I was also thinking that if we don't use the overlay behaviour anywhere, then we should probably just ditch it since it requires just as much effort to size all the images properly to overlay them as it does just to overlay them manually in something like paint.

src/fixate/ui_gui_qt/ui_gui_qt.py
```diff
@@ -266,9 +266,27 @@ def on_image_update(self, path):
...
+            _, ext = os.path.splitext(path)
+            # Default to normal image if not a gif
+            if ext.lower() != ".gif":
+                image = QtGui.QPixmap()
+                image.loadFromData(image_data)
+                self.image_scene.addPixmap(image)
+            else:
+                # NOTE: full overlay features not available with .gif
+                #   due to forcing the scene dimensions to the movie frame
+                if len(self.image_scene.items()) > 0:
```
daniel-montanari, line 278: just use self.image_scene.items()

John2202W: Done

src/fixate/ui_gui_qt/ui_gui_qt.py
John2202W: Created separate user_gif() function.
Have left the image_scene for now and caught some bad cases when switching between gifs and images

PR #148, opened 2022-07-11
Update unit scaling

John2202W:
- Add support for negative values in unit_convert and unit_scale
- Made unit_scale allow case-insensitive unit matching and whitespace
- Fixed possible fall-throughs in unit_convert(): ranges are now hard-coded to eng notation
- Increase test coverage for unit_convert() and unit_scale()
- Add to UNITS scaling options


John2202W:
I've plugged some holes in the unit_convert and by doing so hardcoded the display range to agree with eng.notation. A bunch of different ways this function could be done - depends what we want?

The main intention was to clean up the ui display/input process. e.g. allow the user to enter a value of "6.6uF" when measuring a capacitor value. And from the display side automate the user_info: e.g. "the value should be in the range of 6.5u to 6.7u"

Also - the regex for the unit_scale may be getting nasty and could do with some attempts at breaking it

PR #150, opened 2022-08-02
Sigilent pps driver fixes

John2202W:
Updated regex in sigilent SPD3303X driver to be compatible with latest firmware.

### DIFF
--- a/src/fixate/drivers/pps/siglent_spd_3303X.py
+++ b/src/fixate/drivers/pps/siglent_spd_3303X.py
@@ -220,8 +220,7 @@ class SPD3303X(PPS):
                 return [(code, msg)]
             else:
                 raise InstrumentError(
-                    "Error Returned from PPS\n"
-                    + f"Code: {code}\nMessage: {msg}"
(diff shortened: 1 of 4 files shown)


src/fixate/drivers/pps/siglent_spd_3303X.py
```diff
@@ -206,11 +205,10 @@ def _write(self, data):
 
     def _check_errors(self):
         resp = self.instrument.query("SYST:ERR?")
-        comp = re.compile(r"(\d+) {1,2}([\w ]+)", re.UNICODE)
+        comp = re.compile(r"([+-]?\d+)\,? {1,2}([\w ]+)", re.UNICODE)
         resp = comp.match(resp)
```
clint-lawrence, line 209: Would be good a have a few example strings you're trying to match against.
Or even better - a couple of unittests. If you add unit tests, probably easy to move the query out of the function.

Actually, it might even be worth cleaning up _is_error, since it is never called with the silent keyword arg. Just get rid of that logic altogether and simplify everything?

John2202W: Done all

PR #152, opened 2022-08-18
Update driver imports

John2202W:
Update the driver imports to try opening all matching visa instruments until one is successfully connected.

Fixes current issue where it only attempts to open the first instrument of the desired type in the local_config.

**NOTE**: only applied to the dmm for now. Would end up removing find_instrument_by_id from the config/__init__.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ venv/
 venv
 venv37
 .venv
+.vscode/settings.json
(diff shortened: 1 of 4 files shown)


John2202W:
@clint-lawrence thoughts?

John2202W:
Decision was to not implement this. Closing  PR.

PR #153, opened 2022-08-26
Relocate config/log files, rename local_config.json

John2202W:
Relocate config/log files from the local "fixate" directory to site_config_data given from platformdirs.
i.e. platformdirs.site_config_dir("fixate", False)
On Win10 = "C:\ProgramData\fixate".

Rename local_config.json -> instruments.json


src/fixate/__main__.py
```diff
@@ -364,7 +364,9 @@ def run_main_program(test_script_path=None):
 
     args, unknown = parser.parse_known_args()
     if not args.disable_logs:
-        handler = RotateEachInstanceHandler("fixate.log", backupCount=10)
+        handler = RotateEachInstanceHandler(
+            fixate.config.CONFIG_DIRECTORY / "fixate.log", backupCount=10
+        )
         handler.setFormatter(
```
clint-lawrence, line 370: `platformdirs.user_log_dir()`?
Generally good to keeps logs separate so you can easily cleanup/delete logs without risking deleting config. (I think...)

John2202W: Done

src/fixate/config/__init__.py
```diff
@@ -20,13 +18,28 @@
...
+LOG_DIRECTORY = Path(platformdirs.user_log_dir("Fixate", False))
+INSTRUMENT_CONFIG_FILE = CONFIG_DIRECTORY / "instruments.json"
 
-LOCAL_CONFIG_PATH = os.path.join(os.path.dirname(__file__), "local_config.json")
+if not CONFIG_DIRECTORY.exists():
+    # Create directory if it doesn't already exist
+    CONFIG_DIRECTORY.mkdir(parents=True, exist_ok=True)
+
+if not LOG_DIRECTORY.exists():
+    LOG_DIRECTORY.mkdir(parents=True, exist_ok=True)
```
clint-lawrence, line 37: I'd like to avoid any logic running at import time. The log directory check should move into the `if not args.disable_logs:` block in __main__

The CONFIG_DIRECTORY check probably can move to fxconfig? that the only place we attempt to write out a file to that location, I think...

clint-lawrence: I note that there are probably a bunch of counter examples still littered through this code base :)

clint-lawrence (changes requested):

src/fixate/config/__init__.py
John2202W: Done now

PR #156, opened 2022-10-20
Add 3.10 & 3.11 to github actions

clint-lawrence:


John2202W:
Worth adding to the tox envs aswell to catch local testing?

PR #157, opened 2022-10-20
Update release notes

John2202W:
Update release notes and uprev version.


John2202W:
@jcollins1983  - should be good for release with this now

PR #158, opened 2022-10-22
Update versions of actions dependencies

jcollins1983:
Remove test PyPI and update action versions in step with test.yml

### DIFF
--- a/.github/workflows/upload_pypi.yml
+++ b/.github/workflows/upload_pypi.yml
@@ -32,6 +32,12 @@ jobs:
     - name: Build the package
       run: |
         python -m build --sdist --wheel --outdir dist/ .
+    - name: Publish distribution to Test PyPI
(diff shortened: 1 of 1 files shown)


clint-lawrence:
I assume you're happy to just close this now that test pypi is back up?

jcollins1983:
We could, but do we really need the test PyPI in the workflow? I suppose if something fails there, we won't be forced to up rev on the real PyPI. If we don't delete the test PyPI, wouldn't we still want to update the actions?

pazzarpj:
I guess I didn't really get it sorted in the beginning.  But the intention
of the test pypi would be to check that all the artifacts are there by say
doing a fresh pip install from test pypi and running the tests.

jcollins1983:
Reinstated test PyPI, only changes now are to the action versions to be inline with those in test.yml

PR #160, opened 2022-10-24
Fix local pytest log checking

John2202W:
Overwrite any potential local_config (fixate.yml) so pytests can be run locally.

Since the change in config file directories - local fixate.yml lives and persists outside the venv. This causes it to override the default logging format which the test_script2log expects. Therefore when tests are run locally they will fail when fixate.yml differs from the default.

Two possible options:
1. Re-overwrite the fixate.yml with the default (quick/dirty solution - I have done here initially)
2. Remove the fixate.yml as a 'global' config that is always used and instead pass it independently through the --config command line arg.

Since it is a subprocess call I don't believe we can patch something into fixate.

### DIFF
--- a/test/core/expect-logs/fixate.yml
+++ b/test/core/expect-logs/fixate.yml
@@ -1,0 +1,14 @@
+tpl_csv_path: "{start_date_time}-{index}.csv"
+tpl_time_stamp: '{0:%Y}{0:%m}{0:%d}-{0:%H}{0:%M}{0:%S}'
+
+plg_csv:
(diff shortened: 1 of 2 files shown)


clint-lawrence:
I think what you've implemented makes sense. Have the test control exactly what it needs and pass through as a cli option when calling fixate as a subprocess.

jcollins1983:
> I think what you've implemented makes sense. Have the test control exactly what it needs and pass through as a cli option when calling fixate as a subprocess.

No objections to me merging this PR then? Looks good to me.

PR #162, opened 2022-11-14
Keithly dmm driver addition

Jasper-Harvey0:
I have the driver for the new DMM working. 
Cleaned up some unused code in the Fluke driver too

### DIFF
--- a/src/fixate/drivers/dmm/__init__.py
+++ b/src/fixate/drivers/dmm/__init__.py
@@ -11,16 +11,23 @@ import pyvisa
 import fixate.drivers
 from fixate.drivers.dmm.helper import DMM
 from fixate.drivers.dmm.fluke_8846a import Fluke8846A
+from fixate.drivers.dmm.keithley_6500 import Keithley6500
 from fixate.config import find_instrument_by_id
 
 
 def open() -> DMM:
-    instrument = find_instrument_by_id(Fluke8846A.REGEX_ID)
-    if instrument is not None:
-        # we've found a connected instrument so open and return it
-        rm = pyvisa.ResourceManager()
-        # open_resource could raise visa.VisaIOError?
-        driver = Fluke8846A(rm.open_resource(instrument.address))
-        fixate.drivers.log_instrument_open(driver)
-        return driver
+    # for ID in DMM_list:
+    for DMM in [Fluke8846A, Keithley6500]:
+        instrument = find_instrument_by_id(DMM.REGEX_ID)
+        if instrument is not None:
+            # we've found a connected instrument so open and return it
+            rm = pyvisa.ResourceManager()
+            # open_resource could raise visa.VisaIOError?
+            try:
+                driver = DMM(rm.open_resource(instrument.address))
+                fixate.drivers.log_instrument_open(driver)
+                return driver
+            except pyvisa.errors.VisaIOError:
+                # Tried to open a DMM that is in the config, but not physically connected.
+                pass
     raise fixate.drivers.InstrumentNotFoundError
--- a/src/fixate/drivers/dmm/__init__.py
+++ b/src/fixate/drivers/dmm/__init__.py
@@ -16,13 +16,12 @@ from fixate.config import find_instrument_by_id
 
 
 def open() -> DMM:
-    # for ID in DMM_list:
+
     for DMM in [Fluke8846A, Keithley6500]:
         instrument = find_instrument_by_id(DMM.REGEX_ID)
         if instrument is not None:
-            # we've found a connected instrument so open and return it
+            # Attempt to open instrument in config file
             rm = pyvisa.ResourceManager()
-            # open_resource could raise visa.VisaIOError?
             try:
                 driver = DMM(rm.open_resource(instrument.address))
                 fixate.drivers.log_instrument_open(driver)
--- a/src/fixate/drivers/dmm/__init__.py
+++ b/src/fixate/drivers/dmm/__init__.py
@@ -13,22 +13,26 @@ import fixate.drivers
 from fixate.config import find_instrument_by_id
 from fixate.drivers import InstrumentNotFoundError, InstrumentOpenError
 from fixate.drivers.dmm.fluke_8846a import Fluke8846A
+from fixate.drivers.dmm.keithley_6500 import Keithley6500
 from fixate.drivers.dmm.helper import DMM
 
 
 def open() -> DMM:
-    instrument = find_instrument_by_id(Fluke8846A.REGEX_ID)
-    if instrument is not None:
-        # We've found a configured instrument so try to open it
-        rm = pyvisa.ResourceManager()
-        try:
-            resource = rm.open_resource(instrument.address)
-        except pyvisa.VisaIOError as e:
-            raise InstrumentOpenError(
-                f"Unable to open DMM: {instrument.address}"
-            ) from e
-        # Instantiate driver with connected instrument
-        driver = Fluke8846A(resource)
-        fixate.drivers.log_instrument_open(driver)
-        return driver
+
+    for DMM in (Fluke8846A, Keithley6500):
+
+        instrument = find_instrument_by_id(DMM.REGEX_ID)
+        if instrument is not None:
+            # We've found a configured instrument so try to open it
+            rm = pyvisa.ResourceManager()
+            try:
+                resource = rm.open_resource(instrument.address)
+            except pyvisa.VisaIOError as e:
+                raise InstrumentOpenError(
+                    f"Unable to open DMM: {instrument.address}"
+                ) from e
+            # Instantiate driver with connected instrument
+            driver = DMM(resource)
+            fixate.drivers.log_instrument_open(driver)
+            return driver
     raise InstrumentNotFoundError
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -73,7 +73,7 @@ class Fluke8846A(DMM):
 
     @samples.setter
     def samples(self, val):
-        self._write(["SAMP:COUN {}".format(self.samples)])
+        self._write(f"SAMP:COUN {val}")
         self._is_error()
         self._samples = val
 
@@ -84,14 +84,19 @@ class Fluke8846A(DMM):
         self._write("SYST:REM")
 
     def set_manual_trigger(self, samples=1):
+        MAX_TRIGGER_COUNT = 5000
         self._manual_trigger = True
         self.samples = samples
-        self._write("TRIG:SOUR BUS")  # set DMM to remote trigger
-        self._write("TRIG:COUN {}".format(samples))  # Set number of samples
+        # set DMM to remote trigger
+        self._write("TRIG:SOUR BUS")
+        # Set number of samples to maximum:
+        self._write(f"TRIG:COUN {int(MAX_TRIGGER_COUNT/samples)}")
         self._write("INIT")  # Wait for trigger
+        self._is_error()  # Catch possible insufficient memory error (and others)
 
     def trigger(self):
         self._write("*TRG")  # Send trigger to instrument
+        self._is_error()  # Catch errors. This might slow things down
 
     def measurement(self):
         """
@@ -113,7 +118,7 @@ class Fluke8846A(DMM):
         """
         with self.lock:
             self._is_error(silent=True)
-            self._write(["*rst", "SYST:REM", "*cls", "disp {}".format(self.display)])
+            self._write(["*rst", "SYST:REM", "*cls", f"disp {self.display}"])
             self._CLEAN_UP_FLAG = False
             self._is_error()
 
@@ -151,6 +156,9 @@ class Fluke8846A(DMM):
         """
         if self._manual_trigger:
             values = self.instrument.query_ascii_values("FETCH?")
+            # Reset for next set of measurements (clear buffer). 
+            # Fluke does not allow you to manually clear the buffer, so this roundabout way is used instead
+            self.set_manual_trigger(samples=self.samples)
         else:
             values = self.instrument.query_ascii_values("READ?")
 
@@ -194,7 +202,7 @@ class Fluke8846A(DMM):
                     "Error(s) Returned from DMM\n"
                     + "\n".join(
                         [
-                            "Code: {}\nMessage:{}".format(code, msg)
+                            "Code: {code}\nMessage:{msg}"
                             for code, msg in errors
                         ]
                     )
@@ -225,11 +233,11 @@ class Fluke8846A(DMM):
         """
         self.mode = mode
         self._manual_trigger = False  # Default mode is auto trigger
-        mode_str = "{}".format(self._modes[self._mode])
+        mode_str = f"{self._modes[self._mode]}"
         if _range is not None:
-            mode_str += " {}".format(_range)
+            mode_str += f" {_range}"
         if suffix is not None:
-            mode_str += " {}".format(suffix)
+            mode_str += f" {suffix}"
         self._write(mode_str)
         self._write(
             [
@@ -237,7 +245,7 @@ class Fluke8846A(DMM):
                 "TRIG:DEL:AUTO ON",
                 "TRIG:SOUR IMM",
                 "TRIG:COUN 1",
-                "SAMP:COUN {}".format(self.samples),
+                f"SAMP:COUN {self.samples}",
             ]
         )
         self._is_error()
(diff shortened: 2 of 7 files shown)


src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,299 @@
...
+        Attempts to read values from the DMM up until self.retrys_on_timeout amount of times
+        After each attempt DMM is checked for errors
+        If errors.VisaIOError during read error_cleanup is called
+        raise: VisaIOError if exception on the last attempt
+               ValueError if no values are read
+        return: values read from the DMM
+        """
+        self._write(
+            ["TRAC:CLE"]
+        )  # Clear the reading buffer for next set of measurements.
```
John2202W, line 141: Can clean up these black formatting changes by manually moving the comment

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,299 @@
...
+    def period(self, _range=None):
+        # Cannot set range for period measurement
+        self._set_measurement_mode("period")
+
+    def capacitance(self, _range=None):
+        self._set_measurement_mode("capacitance", _range)
+
+    def diode(self, low_current=True, high_voltage=False):
+        # Cannot set range. 10V fixed range
+        self._set_measurement_mode("diode")
```
John2202W, line 258: Is the single keithley diode test equivalent to the flukes high_current + high_voltage? There's a few zener tests in test scripts that rely on that.
Different current modes may also affect diode Vf test values?

Jasper-Harvey0: I have fixed this up. Found a command to set the bias current, but I think the range (high/low voltage) setting has to remain fixed. 
I assume its primarily the bias current that matters the most? Testing on the jig I have gives better measurements now.

John2202W: If each modes bias currents are the same and the keithley's default voltage is ≥ the fluke's when high_voltage=True then it should be fine

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,302 @@
...
+    def samples(self, val):
+        # Sample number is per mode.
+
+        # Clip values to upper and lower bounds. DMM likes to crash if set out of bounds
+        if val < 1:
+            val = 1
+        elif val > 1000000:
+            val = 1000000
+
+        self._write(["SENS:COUN {}".format(val)])
```
John2202W, line 78: Minor comments that I can see came from the old driver:
- Since there's an instance check in _write: a single string doesn't need to be in a list. 
- f-strings can replace .format

Jasper-Harvey0:
I am still hunting some issues where the DMM will crash after being given a bad command, requiring a power cycle.

jcollins1983 (changes requested):
There are a couple of things that need tidying up, @John2202W picked up some good things.

A general comment: There is a LOT of duplicated code between the two drivers, but how many more DMM drivers are we going to write? So I'm reluctant to ask for a refactor to pull out the commonality. It feels like an abstract base class would be appropriate here as an alternate to the "interface" that's defined in the DMM class.

src/fixate/drivers/dmm/__init__.py
```diff
@@ -11,16 +11,22 @@
...
+        if instrument is not None:
+            # Attempt to open instrument in config file
+            rm = pyvisa.ResourceManager()
+            try:
+                driver = DMM(rm.open_resource(instrument.address))
+                fixate.drivers.log_instrument_open(driver)
+                return driver
+            except pyvisa.errors.VisaIOError:
+                # Tried to open a DMM that is in the config, but not physically connected.
+                pass
```
John2202W, line 31: There are two different errors here that we probably want to differentiate: an instrument exists in the config but can't be opened (pyvisa error), or there is no instrument configured on the station (InstrumentNotFoundError).

In the visa except - should at least 'break' instead of pass as the consensus was not allowing multiple devices of the same type to be configured/tried.

I started PR #161 to better handle the instrument open errors - if/when merged could consolidate

jcollins1983: I think I'll get PR #161 merged, then we can look at what changes need to be made here taking the changes there into account.

clint-lawrence: If this is otherwise ready to merge, you could add a log in the except clause, so the issues is not completely silently ignored. Then #161 can tackle the wider issue.

jcollins1983: You can't tell from the snippet above, but in the wider context the VisaIOError here is being used for flow control (not a huge fan of using exceptions for flow control, but I'm assuming open_resource doesn't just return None if it couldn't find the instrument) for the loop through the DMM list. If nothing found, InstrumentNotFoundError will still be raised, so I think this instance is one where silent ignorance is ok.

As I type this, I decided I do agree with adding a log message (and possibly display a warning message) that there is an instrument in the config that's not connected (is my memory correct in that we don't really want instruments hanging around in the config if they're not connected?)

jcollins1983:
Also, @Jasper-Harvey0, can you please update the release notes before you put the PR back to open?

Jasper-Harvey0:
I have made a few updates to the drivers to help remove direct writes to the instruments in some test scripts.

Main change is functions for manual triggering of the DMM. Typical test scripts that use manual triggering look something like: 
`sample_count = 7
dmm._write("TRIG:SOUR BUS")                  # set DMM to remote trigger
dmm._write(f"TRIG:COUN {sample_count}")      # set amount of tigger counts   
time.sleep(1)

dmm._write("INIT")                       # Arm DMM for next measurement
for trig in range(0, sample_count):      # loop manually to collect samples and step voltage
    dmm._write("*TRG")                   # trigger DMM measurement
    if trig == 5:                        # voltage step change part way through loop
        # Do stuff
        pass
samples = dmm.instrument.query_ascii_values("FETCH?")    # retrieve measurements from DMM`

This now looks like the following:
`sample_count = 7 
dmm.set_manual_trigger(samples=1) # Set to manual trigger mode. 1 sample per trigger event

for trig in range(0, sample_count):
    dmm._write("*TRG")
    if trig == 5:
        # Do stuff
        pass
samples = dmm.measurements() # Collect measurements from buffer`

src/fixate/drivers/dmm/__init__.py
Jasper-Harvey0: My assumption when first writing that was that the config can have both the Keithley and Fluke instruments set up. I have since changed that to assume only one instrument is present in the config (the way it should be?).

src/fixate/drivers/dmm/keithley_6500.py
Jasper-Harvey0: Diode bias current behavior should be normalized between the two drivers.

Jasper-Harvey0:
@jcollins1983 Can we get this merged?

src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -5,38 +5,6 @@
...
-    ":CURRent": {
-        ":DC": {" [{range}]": {", [{resolution}]": {}}},
-        ":AC": {" [{range}]": {", [{resolution}]": {}}},
-    },
-    ":RESistance": {" [{range}]": {", [{resolution}]": {}}},
-    ":FRESistance": {" [{range}]": {", [{resolution}]": {}}},
-    ":FREQuency": {" [{range}]": {", [{resolution}]": {}}},
-    ":PERiod": {" [{range}]": {", [{resolution}]": {}}},
-    ":CAPacitance": {" [{range}]": {", [{resolution}]": {}}},
-    ":TEMPerature": {":FRTD": {" [{RTD_type}]": {}}, ":RTD": {" [{RTD_type}]": {}}},
```
John2202W, line 22: Somewhere along the line we've lost the temperature measurement in this driver, may not be important as it was likely never used

John2202W (commented):
Couple of extra comments. Got to decide whether we want to support analog/digital_filter() and temperature measurements, if so should be on both drivers,
error_cleanup() in the Fluke could also be useful (currently not used anywhere either)

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,343 @@
...
+            return self._read_measurements()
+
+    def reset(self):
+        """
+        Checks for errors and then returns DMM to power up state
+        """
+        with self.lock:
+            self._is_error(silent=True)
+            # Wait for previous commands to finish, reset, clear event logs
+            # self._write("*WAI; *RST; *WAI; *CLS")
```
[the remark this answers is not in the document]
Jasper-Harvey0: Must have commented out to hunt for what was causing DMM crashes...

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,343 @@
...
+        self._display = "ON100"
+        # del self.instrument.timeout
+        self.instrument.timeout = 10000
+        self.instrument.query_delay = 0.3  # Stop DMM crash
+        self.instrument.delay = 0  # Stop DMM crash
+        self.is_connected = True
+        self.reset()
+        self._manual_trigger = False
+        self._samples = 1
+        self._CLEAN_UP_FLAG = False
```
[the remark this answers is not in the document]
Jasper-Harvey0: I guess I kept a lot of these flags just in case some peoples code go looking for them? In the end I guess its a problem with not having a proper base class.

Yes, I neglected the analog and digital filters because they aren't used, and I don't really see the benefit for having them. The Keithley does not have an analog filter either.

src/fixate/drivers/dmm/fluke_8846a.py
Jasper-Harvey0: Yes. Temperature was neglected because it's not used.

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,343 @@
...
+        self.samples = samples
+        self._write("TRIG:LOAD 'EMPTY'")  # Load empty model
+        self._write(f"TRIG:BLOC:MDIG 1, 'defbuffer1', {samples}")
+        self._write("TRAC:CLE")
+
+    def trigger(self):
+        """
+        Manually trigger a measurement and store in instrument buffer.
+        """
+        self._write("INIT; *WAI")
```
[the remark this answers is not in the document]
Jasper-Harvey0: This will now throw an InstrumentError exception. Let me know what the preference for handling this is.

jcollins1983:
@Jasper-Harvey0, you will need to resolve the conflict on the releae notes now.

src/fixate/drivers/dmm/keithley_6500.py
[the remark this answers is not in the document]
jcollins1983: The atgrep tool doesn't return any hits for _CLEAN_UP_FLAG, _ANALOG_FLAG or _DIGITAL_FLAG. I'm not sure why any of them exist. I'd be inclined to remove them, we can always add them back in later if there ever is a need for them.
For the stuff in the helper, I'm not sure which side of the fence that I stand on for requiring all of the functions in the helper file to be implemented. We definitely need any method in a driver to appear in the helper because of how fixate works, but it's possible that we might have another DMM that doesn't support other methods that are declared in the helper. So we might need to think about how we deal with the not-implemented errors that will occur if someone uses a method that's not supported on a particular instrument.

src/fixate/drivers/dmm/keithley_6500.py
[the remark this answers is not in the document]
jcollins1983: If trigger will only ever be called for manual triggering (which I imagine is the case) then we could call set_manual_trigger first. We could use `self._manual_trigger` to determine whether it needs to be run or not and just use the default values if `set_manual_trigger` hasn't been called. 
What is the usecase for manual triggering? atgrep doesn't return any hits on set_manual_trigger or trigger.

src/fixate/drivers/dmm/keithley_6500.py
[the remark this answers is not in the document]
jcollins1983: Update, the fluke driver sets `_CLEAN_UP_FLAG` in `error_cleanup()`, but this isn't done in this driver.

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def _write(self, data):
+        """
+        Writes data to the DMM
+        raise: ParameterError if called with no data string
+        """
+        if data:
+            if isinstance(data, str):
+                self.instrument.write(data)
+                time.sleep(0.05)
```
jcollins1983, line 158: I'm not a fan on uncommented sleeps. If we need the sleep for a reason it should be mentioned here.

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -0,0 +1,340 @@
...
+            else:
+                for itm in data:
+                    self.instrument.write(itm)
```
jcollins1983, line 161: There's a sleep after the write above, but not this one? We only sleep at the end of the loop. Is this intentional? Makes me wonder even more if the previous sleep is even needed.
Also, we've landed in where because `data` wasn't a string, but what if it's an in or list of ints, or anything else? I imagine we're wanting it to to be a list of strings, what gets raised if it gets something that's not?
Perhaps the below would be more robust?
```
if not isinstance(data, str) and not all([isinstance(itm, str) for itm in data]):
    raise ParameterError("...")
```

jcollins1983: That would need a bit more thought put into it, because it will break in some cases. But I think that general idea would be more robust.

src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -105,10 +73,31 @@ def samples(self):
...
+        self.samples = samples
+        # set DMM to remote trigger
+        self._write("TRIG:SOUR BUS")
+        # Set number of samples to maximum:
+        self._write(f"TRIG:COUN {int(MAX_TRIGGER_COUNT/samples)}")
+        self._write("INIT")  # Wait for trigger
+        self._is_error()  # Catch possible insufficient memory error (and others)
+
+    def trigger(self):
+        self._write("*TRG")  # Send trigger to instrument
```
jcollins1983, line 98: The same comment from the keithly driver applies here.

src/fixate/drivers/dmm/keithley_6500.py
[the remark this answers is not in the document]
Jasper-Harvey0: The reason for adding manual triggering was to remove manual writes to the instruments. There are a few places in test scripts that we have manual writing happening to trigger the DMM. 
atgrep *TRG will show where this is happening. 

Changing this to run the set trigger is probably fine. I guess I wanted to encourage people to be mindful of how many samples they want to take. Forcing them to set the trigger mode is one way to help encourage that.

src/fixate/drivers/dmm/keithley_6500.py
Jasper-Harvey0: Do we have issues with robustness at the moment for this? The only way this function breaks is if something calling it internally sends bad data, seeing as the _write function is not meant to be used externally. 

If you really want this changed, then that's fine, I just think we're fighting small battles here.

src/fixate/drivers/dmm/keithley_6500.py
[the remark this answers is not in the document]
John2202W: Could add to trigger() - to give an informative message
```
if not self._manual_trigger:
    raise InstrumentError("Manual trigger mode not set")
```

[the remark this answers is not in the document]
jcollins1983: I was in the process of typing the same thing.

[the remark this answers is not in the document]
jcollins1983: > The reason for adding manual triggering was to remove manual writes to the instruments. There are a few places in test scripts that we have manual writing happening to trigger the DMM. atgrep *TRG will show where this is happening.
> 
> Changing this to run the set trigger is probably fine. I guess I wanted to encourage people to be mindful of how many samples they want to take. Forcing them to set the trigger mode is one way to help encourage that.

Also, I didn't notice the comment [here](https://github.com/PyFixate/Fixate/pull/162#issuecomment-1324498900) that told me the same thing.

src/fixate/drivers/dmm/keithley_6500.py
Jasper-Harvey0: Now raises a ParameterError if data is not str or list of str

src/fixate/drivers/dmm/keithley_6500.py
Jasper-Harvey0: Sleeps commented

src/fixate/drivers/dmm/keithley_6500.py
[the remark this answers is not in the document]
Jasper-Harvey0: Clean up flags are removed, as well as unused clean up functions

jcollins1983 (approved):

PR #145, opened 2022-07-04
Fixate Driver Tests

Jasper-Harvey0:
Made a start on the driver tests. 

Definitely going to be some silly stuff in here that needs changing

PR #164, opened 2022-11-24
Driver Tests

Jasper-Harvey0:
Git is hard... Had to make a branch and re-commit all these files.

This is a set of tests for the instrument drivers.


jcollins1983:
There were a bunch of comments and review on the original PR #145, as @Jasper-Harvey0 said, git is hard. But we both learnt some stuff.

PR #169, opened 2023-04-21
Change default returncode to Error

John2202W:
Change default returncode to Error - only way it should pass is by finishing under correct conditions. 
(more details in CPE-2586)

Delete unused sequencer variable.

PR #170, opened 2023-04-26
Add logging of ftdi instruments

John2202W:
Add logging of FTDI instruments.
Allow substring matching to the ftdi description.
(more details - CPE2543)

### DIFF
--- a/src/fixate/drivers/ftdi.py
+++ b/src/fixate/drivers/ftdi.py
@@ -116,7 +116,7 @@ if os.name == "nt":
         ftdI2xx = ctypes.WinDLL("FTD2XX.dll")
     except Exception as e:
         raise ImportError(
-            "Unable to find FTD2XX.dll.\nPlugging in FDTI device will install DLL."
+            "Unable to find FTD2XX.dll.\nPlugging in an FTDI device will install the DLL."
         ) from e
 else:
     try:
--- a/src/fixate/drivers/ftdi.py
+++ b/src/fixate/drivers/ftdi.py
@@ -2,6 +2,7 @@ import ctypes
 import struct
 import time
 import os
+import re
 
 import fixate.drivers
 from fixate.core.common import bits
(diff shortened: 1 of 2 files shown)


src/fixate/drivers/ftdi.py
```diff
@@ -450,3 +441,21 @@ def _serial_shift_bit_bang(self, data, bytes_required, bb_mask):
         data_out.append(bb_mask + self.bb_latch ^ self.bb_inv_mask)
         data_out.append(bb_mask + self.bb_inv_mask)
         return data_out
+
+    def get_identity(self) -> str:
+        """Return identity string representing connected ftdi object"""
+        return f"{self.ftdi_description.decode()}"
+
```
clint-lawrence, line 448: `return self.ftdi_description.decode()`?

John2202W: The ftdi_description comes from the raw FT Device info. Otherwise ends with the raw string as b'InstrumentX'

John2202W: Yes, previously had some some other details in there but can fix that now

John2202W: Done

clint-lawrence:
I don't anticipate this being a problem, but I occurred to me that you can't prevent a substring matching. For example if we had two devices, "abc" and "abcd", and you really wanted to only open "abc", not "abcd" there is no way to specify that. I can't imagine a realistic scenario where that would come up.

Could use fnmatch to add wildcard matching or a regex if we need to.

John2202W:
Have changed to regex to give more flexibility if/when needed. Still up to the test-script author to prevent unwanted substrings
e.g `ftdi.open(r"abc\-?\d? Some Description$")`

clint-lawrence:
What about using `re.fullmatch` instead of `re.match`? With that the existing behavior of strict matching will be maintained, but allows more flexible matching going forward.

John2202W:
Updated ^

clint-lawrence:
Looks good to me

jcollins1983:
Looks good. I can merge once the conflict on the release notes in resolved from the merge of PR #169

PR #179, opened 2024-01-08
Add computername logging

Jasper-Harvey0:
Recently I have been wanting to filter fixate logs by asset number to get an idea if a specific amptower has been failing more than others. This should allow us to do this. 

I am not sure how to write a test for this, as the computername string will change every time fixate is run on a new computer. Maybe someone has as idea?

### DIFF
--- a/src/fixate/config/__init__.py
+++ b/src/fixate/config/__init__.py
@@ -12,6 +12,8 @@ from fixate.config.helper import (
     get_config_dict,
     render_template,
 )
+
+import platform
 import os.path
 import json
 import dataclasses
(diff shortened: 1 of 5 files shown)


src/fixate/config/__init__.py
```diff
@@ -28,6 +28,13 @@
 LOG_DIRECTORY = Path(platformdirs.user_log_dir("Fixate", False))
 INSTRUMENT_CONFIG_FILE = CONFIG_DIRECTORY / "instruments.json"
 
+try:
```
daniel-montanari, line 31: platform.node() and platform.uname() look like a better way to get this type of info without having to deal with cross platform issues of the os module.
https://docs.python.org/3/library/platform.html#platform.node

Jasper-Harvey0: Yep, thanks!

clint-lawrence:
@Jasper-Harvey0 to test, you could run a test script, look for the computer name log entry, then check is matches `platform.node`?

src/fixate/config/__init__.py
```diff
@@ -51,6 +54,7 @@
         "test-script-name={test_script_name}",
         "report-format={REPORT_FORMAT_VERSION}",
         "index_string={index}",
+        "computername={COMPUTERNAME}",
```
John2202W, line 57: Just noting in case you haven't followed it through before but this default config is overwritten when called through amptest. So will need to be updated in the src there too,

John2202W:
Just noting in case you haven't followed it through before but the default reporting config is overwritten when called through amptest. So will need to be updated in the src there too in the future to release

Jasper-Harvey0:
Had to add another yaml log config to leave the current tests unaffected. 

Thanks for that John, I did run into that issue when doing my testing.

clint-lawrence (approved):

PR #181, opened 2024-03-26
Dmm driver improvements

Jasper-Harvey0:
Add delay option to DMM driver to help alleviate issues with Keithley DMM compatibility. 
This stems from test script failures due to the Keithley sampling a measurement before the relays (or voltage rails etc) have completely settled. Apparently the Fluke is slower at this so it hasn't been a problem. 

I also removed some dead code and formatted.

### DIFF
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -9,9 +9,8 @@ class Fluke8846A(DMM):
     INSTR_TYPE = "VISA"
 
     def __init__(self, instrument, *args, **kwargs):
-        self.measurement_delay = (
-            0.2  # Delay between call to self.measurement() and querying the DMM.
-        )
+        # Delay between call to self.measurement() and querying the DMM.
+        self.measurement_delay = 0
         self.instrument = instrument
         instrument.rtscts = 1
         self.lock = Lock()
@@ -81,15 +80,18 @@ class Fluke8846A(DMM):
         self._write("*TRG")  # Send trigger to instrument
         self._is_error()  # Catch errors. This might slow things down
 
-    def measurement(self, delay=False):
+    def measurement(self, delay=None):
         """
         Sets up DMM triggering, creates list of measurements from the read buffer
 
-        delay: If True, will wait for self.measurement_delay seconds before triggering a measurement.
+        delay: If not set, will wait for self.measurement_delay seconds before triggering a measurement. If set, will wait for delay seconds before triggering a measurement.
         returns: a single value as a float
         """
-        if delay:
-            time.sleep(self.measurement_delay)
+        if delay is None:
+            delay = self.measurement_delay
+
+        if delay > 0:
+            time.sleep(delay)
         return self.measurements()[0]
 
     def measurements(self):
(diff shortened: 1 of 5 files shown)


src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -100,11 +81,15 @@ def trigger(self):
...
-    def measurement(self):
+    def measurement(self, delay=False):
         """
         Sets up DMM triggering, creates list of measurements from the read buffer
+
+        delay: If True, will wait for self.measurement_delay seconds before triggering a measurement.
         returns: a single value as a float
         """
+        if delay:
+            time.sleep(self.measurement_delay)
```
clint-lawrence, line 92: how about this. Then you can override the default for a single measurement. And if you want a delay on every measurement, you don't have to pass in `delay = True` everywhere.
```python
def measurement(self, delay=None):
    if delay is None:
        delay = self.measurement_delay
    
    if delay > 0:
        time.sleep(delay)
````

File: src/fixate/drivers/dmm/fluke_8846a.py [hunk 2]
```diff
@@ -11,6 +9,9 @@ class Fluke8846A(DMM):
     INSTR_TYPE = "VISA"
 
     def __init__(self, instrument, *args, **kwargs):
+        self.measurement_delay = (
+            0.2  # Delay between call to self.measurement() and querying the DMM.
```
clint-lawrence, line 13: Can you pull the comment up to the line above to avoid the assignment being split across lines.

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -11,6 +9,9 @@ class Keithley6500(DMM):
     INSTR_TYPE = "VISA"
 
     def __init__(self, instrument, *args, **kwargs):
+        self.measurement_delay = (
+            0.2  # Delay between call to self.measurement() and querying the DMM.
+        )
```
clint-lawrence, line 14: Same formatting issues here

clint-lawrence (changes requested):

src/fixate/drivers/dmm/fluke_8846a.py
Jasper-Harvey0: Good point!

clint-lawrence (approved):
Looks good

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -117,11 +117,18 @@ def trigger(self):
         self._write("INIT; *WAI")
         self._is_error()
 
-    def measurement(self):
+    def measurement(self, delay=True):
```
clint-lawrence, line 120: This should be `delay=None` I think?

src/fixate/drivers/dmm/helper.py
```diff
@@ -28,7 +28,13 @@ def trigger(self):
         """
         raise NotImplementedError
 
-    def measurement(self):
+    def measurement(self, delay=False):
```
clint-lawrence, line 31: `delay=None` ?

clint-lawrence (changes requested):

clint-lawrence (approved):
While I don't think this is the right big picture solution, it is our best pragmatic path forward right now. And the change itself looks good.

PR #182, opened 2024-05-27
Possible driver manager concept

clint-lawrence:
This is pretty scrappy, but mostly trying to test out the big picture idea.

Have a look at `test-script.py`. Here is the general concept:

```python
# 1. Define a class which is your "driver manager"
@dataclasses.dataclass
class Jig123DriverManager:
    dmm: dmm.DMM = dataclasses.field(default_factory=dmm.open)


# 2. TestList and TestClass are generic. You need connect them to the driver manager type
class Jig123TestList(TestList[Jig123DriverManager]):
    pass


class Jig123TestClass(TestClass[Jig123DriverManager]):
    pass

# 3. Define your tests & lists as normal. But now the driver manager will be passed in by the
# sequencer to test, set_up, tear_down, enter and exit. By annotating the dm attribute, you
# get reliable autocomplete in the IDE.
class Test(Jig123TestClass):
    def test(self, dm: Jig123DriverManager):
        value = dm.dmm.measure()
        ...

# 4. Instead of only specifying a top level test list to the sequencer, we will now
# pass in a `TestScript`, which tell the sequencer how to create a driver manager to pass
# into the TestClass/TestList methods.
TEST_SCRIPT = TestScript(
    test_list=Jig123TestList(... Test list goes here ... ),
    dm_type=Jig123DriverManager,
)
```

The doesn't work 100% yet, the example with the "dummy" driver works. obviously we will also need to think about backwards compatibility.

PR #184, opened 2024-06-03
Jig mux alternate implementation

clint-lawrence:
This introduces a new implementation of the jig switching support code. The primary motivation/goals:
- The previous implementation in jig_mapping had tight coupled between the address handlers and virtual mux. This made any changes or refactoring very difficult and made the code harder to understand than needed
- The way JigDriver worked required AddressHandlers to be instantiated when a JigDriver subclass was defined. This more or less forced address handlers to reference a global object for any real IO. 

This implementation make the "pin" concept the primary point of coupling between the AddressHandlers and VirtualMuxs. Then the JigDriver is being revised to improve type hinting of the jig driver object and also decouple instantiation at runtime from JigDriver subclass definition. In addition, I'm working towards the new code have complete (or near complete) type annotations.

While these changes are trying to avoid the need to change existing test scripts, it won't be possible to implement this in a completely backwards compatible way. So this will be introduced as a parallel implementation. Existing scripts will continue to run. New and updated scripts will import from a different submodule and be able to take advantage of the improved implementation. In time, the current jig_mapping implementation will be removed.

Once this work is done, it will be possible to move software with the related DriverManager change https://github.com/PyFixate/Fixate/pull/182/files

- [x]  remove `switch_through_all_signals`. Instead, create a method that returns all signals on a mux.
- [x] modify `iterate_all_mux_paths` to return mux object and signal pairs inline with the change above to `switch_through_all_signals`
- [x] run on some hardware
- [ ] write a migration guide for existing test scripts
- [x] write test for VirtualMux.multiplex
- [x] write test for VirtualAddressMap.add_update
- [x] Are there any additional cross-checks we should implement on virtualmux/addresshandler?
       As a minimum, add a method to address map that ensures all pins from all virtual muxes have pin defined in an address handler. Probably call this during the JigDriver init.
- [x] move imports to fixate.{VirtualMux, JigDriver, ...} etc?
- [x] Decide what to do with `VirtualAddressMap.{update_pin_by_name,update_pins_by_name, __getitem__, __setitem__}`
- [x] Should we allow overriding the default signal `""`? In the previous jig_mapping implementation, the behaviour of the empty signal was hard coded. If you defined some pins against `""`, I'm pretty sure that would be ignored. Should we check that the user script doesn't redefine `""`?
- [x] Do we need a `MuxGroup.force_trigger` or similar, that writes all the current mux states out to hardware, even if the pins haven't changed?
- [x] Do we need a `MuxGroup.trigger()` or `VirtualAddressMap.trigger()` that processes any pin changes in `VirtualAddressMap._pending_updates` (Issues created #189 and #190)
- [x] Change JigDriver to accept a factory for the AddressHandlers? No, add `open()` method to address handlers that Jig Driver will call when it needs to. Make AddressHandlers accept a pin_list in __init__()
- [x] Add a "report" method to JigDriver that prints out all connected muxes and pins from the address handlers? (Issue created #192)
- [x] Add an alias to `RelayMatrixVirtualMux` called `BreakBeforeMakeMux`. (Issue created #191)
- [x] Add a `MakeBeforeBreakMux` subclass of `VirtualMux` (Issue created #191)


src/fixate/core/switching.py
```diff
@@ -0,0 +1,720 @@
...
+    ):
+        if not self.pin_list:
+            self.pin_list = [self.pin_name]
+        super().__init__(update_pins)
+
+
+class RelayMatrixMux(VirtualMux):
+    clearing_time = 0.01
+
+    def _calculate_pins(self, old_signal: Signal, new_signal: Signal) -> tuple[PinSetState, PinSetState]:
```
clint-lawrence, line 421: @jcollins1983 said:
> What's the use case for the old_signal in the _calculate_pins method? It appears to be unused currently.

src/fixate/core/switching.py
```diff
@@ -0,0 +1,720 @@
...
+                              ________   |          |
+        a3_c0----------------|        |--|          |
+        a3_c1----------------| Mux C  |  |          |
+        a3_c2----------------|  4:1   |  |          |
+                     (None)--| x5  x4 |  |          |
+                             |________|  |          |
+                                         |__x1__x0__|
+
+        class Mux(VirtualMux):
+            pin_list = ("x0", "x1", "x2", "x3", "x4")
```
clint-lawrence, line 334: @jcollins1983 said
> Should this example have x5 in the pin list?

src/fixate/core/switching.py
```diff
@@ -0,0 +1,720 @@
...
+
+            mux_c = ("a3_c0", "a3_c1", "a3_c2", None)
+            mux_b = ("a1_b0", "a1_b1", "a1_b2", None)
+
+            map_tree = (
```
clint-lawrence, line 339: @jcollins1983  said
> in the map tree example there should be an entry for a2 based on the diagram right? Or am I missing something?

src/fixate/core/switching.py
```diff
@@ -0,0 +1,720 @@
...
+            "a0",
+            mux_b,
+            shift_nested(mux_c, [2]),  # 2 in indicative on how many pins to skip. This case is (x2, x3) from mux_b
```
clint-lawrence, line 342: @jcollins1983 said
> Will shift_nested be brought in from jig_mapping.py? It looks like the intent is to replace jig_mapping.py with the new switching.py

src/fixate/core/switching.py
```diff
@@ -0,0 +1,720 @@
...
+    #             self.virtual_pin_list,
+    #             bits(
+    #                 self._virtual_pin_values_active,
+    #                 num_bits=len(self.virtual_pin_list),
+    #                 order="LSB",
+    #             ),
+    #         )
+    #     )
+
+    def active_pins(self) -> None:
```
clint-lawrence, line 540: @jcollins1983 said
> I'm not sure what your plan is for the actie_pins method, but I use it when fault finding and debugging, so would be good to keep or at least have an equivalent.

src/fixate/core/switching.py
clint-lawrence: Yeah, I will add a comment for that. If you want to implement make-before-break, I think it is needed. Or at least makes it easier/more obvious. And while the underscore marks it as "private" It is also a method that is intended to be overridden, so I though some level of future proofing was reasonable.

src/fixate/core/switching.py
clint-lawrence: Probably - that is just a copy/paste job from jig_mapping.py

src/fixate/core/switching.py
clint-lawrence: Only if we really need it. It's not used in any jig, only some old tests. If there is a need to keep it, I have a better idea anyway! (But it's a bit more work to implement)

src/fixate/core/switching.py
clint-lawrence: That is good feedback to have. Absolutely will keep anything that is useful, or replace with something better. When all the basics work, I'll have to go back and note everything that has been removed to double check nothing important has been deleted.

clint-lawrence: I've ended up with `active_pins` on the address map and `active_signals` on the mux group
image

I'm open to suggestions on how what the return value is, but I think keeping having those two distinct functions is useful

src/fixate/core/switching.py
clint-lawrence: Added the following
> old_signal isn't currently used, but it is provided in case a future subclass needs to it calculate the update. In particular, it could be useful for make-before-break behaviour.

src/fixate/core/switching.py
clint-lawrence: fixed

src/fixate/core/switching.py
clint-lawrence: Added some comments about this, I'm my alternate idea if it is ever needed

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37,py38,py39,py310,py311,black,mypy
+envlist = py38,py39,py310,py311,py212,black,mypy
```
jcollins1983, line 2: Should this be py312?

clint-lawrence: 🤦‍♂️

clint-lawrence:
New idea. This can probably wait, but I'm thinking of other ways to create a VirtualMux that will be type hint and IDE friendly
```python
# existing
class Mux(VirtualMux):
    pin_list = ("x0", "x1", "x2")
    map_list = (
        ("sig1", "x0"),
        ("sig2", "x1"),
        ("sig3", "x2", "x0"),
    )

#new
@dataclass
class MuxNew(VirtualMux):
    sig1: Sig = signal("x0")
    sig2: Sig = signal("x1")
    sig3: Sig = signal("x2", "x0")

dm.jig.mux.Mux("sig2")
# becomes
dm.jig.mux.NewMux.sig2()
```

Then each signal would be a method on the resulting VirtualMux subclass. This obviously a bigger change. If we implement this, I would probably be an alternative, not something we would necessarily update all existing scripts to use.

Implementation would be something like
```python
Sig = Callable[[VirtualMux, bool], None]

def signal(*pins) -> Sig:
    def _func(self: VirtualMux, trigger_update: bool=False):
        self._do_switch(pins, trigger_update)
    return _func
```

clint-lawrence:
I think that general idea can be made to work with muxs defined as trees.
```python

        #                                   __________
        # a0-------------------------------|          |
        #                       ________   |          |
        # a1_b0----------------|        |--|          |
        # a1_b1----------------| Mux B  |  |          |
        # a1_b2----------------|  4:1   |  |          |
        #              (None)--| x3  x2 |  |          |
        #                      |________|  |          |
        #                                  |  Mux A   |
        # a2-------------------------------|   4:1    |
        #                       ________   |          |
        # a3_c0----------------|        |--|          |
        # a3_c1----------------| Mux C  |  |          |
        # a3_c2----------------|  4:1   |  |          |
        #              (None)--| x5  x4 |  |          |
        #                      |________|  |          |
        #                                  |__x1__x0__|

muxA = MuxTree("x0", "x1")
muxB = MuxTree("x2", "x3", nest=muxA)
muxC = MuxTree("x4", "x5", nest=muxA)

@dataclass
class Mux(VirtualMux):
    a0: Sig = muxA()
    a1_b0: Sig = muxB()
    a1_b1: Sig = muxB()
    a1_b2: Sig = muxB()
    muxB()  # "throw away" one entry, not strictly needed here. But could be if the gap was in a different spot
    a2: Sig = muxA()
    a3_c0: Sig = muxC()
    a3_c1: Sig = muxC()
    a3_c2: Sig = muxC()


# Implementation - Maybe? very not tested...
class MuxTree:
    def __init__(self, *pins: tuple[Pin], nest: MuxTree) -> None:
        self._pins = set(pins)
        self._pin_gen = generate_bit_sets(pins)
        if nest is not None:
            self._fixed_pins = nest().pins
        else:
            self._fixed_pins = set()

    def __call__(self) -> Sig:
        pins = self._fixed_pins  | next(self._pin_gen)
        return signal(*pins)

```

clint-lawrence:
I deleted a few comment and moved the content to #187

src/fixate/drivers/handlers.py
```diff
@@ -0,0 +1,76 @@
...
+    return [f"{relay_matrix_num}K{relay}" for relay in range(1, 17)]
+
+
+# This is a real quick test. Not worth the effort unravelling
+# imports when ftdi isn't importable, just to get this into a
+# proper test right now...
+__expected = (
+    "3K1 3K2 3K3 3K4 3K5 3K6 3K7 3K8 3K9 3K10 3K11 3K12 3K13 3K14 3K15 3K16".split()
+)
+assert _pins_for_one_relay_matrix(3) == __expected
```
daniel-montanari, line 30: Do we want to keep this here or move it to a unit test?

src/fixate/core/switching.py
```diff
@@ -0,0 +1,743 @@
...
+        """
+        Override of _calculate_pins to implement break-before-make switching.
+        """
+        setup = PinSetState(off=self._pin_set, on=frozenset())
+        on_pins = self._signal_map[new_signal]
+        final = PinSetState(off=self._pin_set - on_pins, on=on_pins)
+        return setup, final
+
+
+class AddressHandler:
```
daniel-montanari, line 486: Make this a Protocol if it only exists to document the required methods of a handler?

daniel-montanari: Should all these be moved to handlers.py?

src/fixate/core/switching.py
```diff
@@ -0,0 +1,743 @@
...
+        useful for make-before-break behaviour.
+        """
+        setup = PinSetState()
+        on_pins = self._signal_map[new_signal]
+        final = PinSetState(self._pin_set - on_pins, on_pins)
+        return setup, final
+
+    ###########################################################################
+    # The following methods are intended as implementation detail and
+    # subclasses should avoid overriding.
```
daniel-montanari, line 211: @typing.final can get the type checker to enforce this if we want

src/fixate/core/switching.py
clint-lawrence: > Make this a Protocol if it only exists to document the required methods of a handler?
Yeah, that probably makes sense.

Might be worth looking all the classes once we're happy with the general concept. In my examples at the moment, I've got sub-classing, sub-classing with `@dataclass`, potentially throw protocols in the mix. Hopefully doesn't end up too confusing with all the slight differences...

> Should all these be moved to handlers.py?
I don't think so, but agree it's a bit non-obvious what is going on. The handlers in switching.py don't do any real IO. they are  abstract intended to assist with making useful "real world" handers. The `RelayMatrixHandler` relies on drivers, which is why I moved it there. I think it makes sense for it to live there.

One specific issue I had was that you can't import drivers.ftdi without the driver. After moving `RelayMatrixHandler` in then occurred to me that is might make sense to have concrete vs abstract handers in different place. I'm not too fixed on that idea though

src/fixate/drivers/handlers.py
clint-lawrence: I'm happy to leave it there. It's tiny enough and it was looking like a pain to shuffle things around to get the test running correctly without hardware/drivers so the ftdi module would import.

src/fixate/core/switching.py
clint-lawrence: Except I override it below for the `VirtualSwitch` ;)

jcollins1983 (commented):
Just a couple of questions here and there.

src/fixate/core/switching.py
```diff
@@ -0,0 +1,743 @@
...
+                f"Signal '{signal_output}' not valid for multiplexer '{name}'"
+            )
+
+        setup, final = self._calculate_pins(self._state, signal_output)
+        self._update_pins(PinUpdate(setup, final, self.clearing_time), trigger_update)
+        if signal_output != self._state:
+            self.state_update_time = time.time()
+        self._state = signal_output
+
+    def switch_through_all_signals(self) -> Generator[str, None, None]:
```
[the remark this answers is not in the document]
clint-lawrence: Yeah, I agree. I've created a todo list with this suggestion. And similar change for the JigDriver `iterate_all_mux_paths`

src/fixate/core/switching.py
```diff
@@ -0,0 +1,743 @@
...
+        self._pin_lookup = {
+            pin: bit for pin, bit in zip(self.pin_list, bit_generator())
+        }
+
+    def set_pins(self, pins: Collection[Pin]) -> None:
+        value = sum(self._pin_lookup[pin] for pin in pins)
+        self._update_output(value)
+
+    def _update_output(self, value: int) -> None:
+        # perhaps it's easy to compose by passing the output
```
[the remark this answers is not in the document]
clint-lawrence: I think I'll just leave this as is for now. Now that I've added a built in relay matrix handler, it will be pretty rare that scripts need to implement their own thing anyway.

src/fixate/core/switching.py
```diff
@@ -0,0 +1,743 @@
...
+
+    def close(self) -> None:
+        for handler in self._handlers:
+            handler.close()
+
+    def active_pins(self) -> Set[Pin]:
+        return self.virtual_map.active_pins()
+
+    def reset(self) -> None:
+        """
```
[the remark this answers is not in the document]
clint-lawrence: I'm glad you ask, because I'm not sure and it's worth discussing.

There was a mess before of "default signals" on muxs, "default pins" on address handlers and a mux would implicitly handle an empty string signal as "no pins". `JigDriver.reset()` would set default on address handlers *and* set each mux to it's default signal.

```python
VirtualMux.reset()  # set this mux to signal ""
MuxGroup.reset()  # set all muxes to signal ""
VirtualAddressMap.reset()  # set all pins to off
JigDriver.reset()  # just calls VirtualAddressMap.reset() 
```

My thinking was that `jig.reset()` would be used to reset all pins and `jig.mux.reset()` would be used to reset all muxes. Those two should be equivalent, if every pin is used in a mux. Having said that, I don't know how useful the distinction is and if people writing script would be able to keep the subtle difference in their heads

Perhaps a more useful approach is
```python
jig.reset()  # reset all muxes to signal "" (via the mux group reset method)

# then if a script really needs it
jig.virtual_map.reset()  # would reset all pins, without changing mux states.
```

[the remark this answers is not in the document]
jcollins1983: TBH, I think once all jig scripts were switched over and `jig.reset()` was changed to `jig.mux.reset()` people would copy and paste a similar script and follow the script (pun intended), and not even think about/know about/remember the difference, until someone thinks of a reason to treat the pins and mux states differently and goes looking for a way to do it.

Though I don't mind the specificity of `jig.mux.reset()`, because most of the time that call is made it's with the intent to putting the muxes to their off state.

Do you have any thoughts on this @daniel-montanari ?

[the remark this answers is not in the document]
clint-lawrence: Maybe we could remove `jig.reset()`? Then someone writing the script has to be more conscious about choosing `jig.mux.reset()` or `jig.virtual_map.reset()`. I don't actually like that idea...

I'll wait and see what Dan has to say!

src/fixate/core/switching.py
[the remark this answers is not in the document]
daniel-montanari: I had put these into atgrep and didn't come up with anything

script.py
```diff
@@ -0,0 +1,64 @@
...
+#    class JigMuxGroup(MuxGroup):
+#        NewVirtualMux: _NewVirtualMux
+#    Then the references in the script stay this same.
+#    jig.mux.NewVirtualMux
+# 3. Change the attribute name on mux, like in the example below,
+#    but add some compatibility code to MuxGroup base class so that
+#    attribute lookups that match the Class of one of its attributes
+#    get magically mapped to the correct attribute.
+@dataclass
+class JigMuxGroup(MuxGroup):
```
daniel-montanari, line 56:
```
@dataclass
class JigMuxGroup(MuxGroup):
    MuxOne: 'MuxOne' = field(default_factory=MuxOne)
    mux_two: MuxTwo = field(default_factory=MuxTwo)
    mux_three: MuxThree = field(default_factory=MuxThree)

jig = JigDriver(JigMuxGroup, [Handler()])

jig.mux.MuxOne("sig2", trigger_update=False)
jig.mux.mux_two("sig5")
```
So I've noticed forward annotating this works as long as MuxOne is defined globally. I'm noting this more as a comment that it is possible since personally I'd prefer options 1 and 2.

test/core/test_switching.py
daniel-montanari, line 1: Do we want to add some tests for make a jig and check mux groups work type of things?

script.py
clint-lawrence: Interesting that that works, but I'm not sure what you mean by "defined globally", 

Also worth noting that we don't have to do the same thing in every script. (But having a convention is a good thing)

Anyway - I'm squarely on Team Option 1 and I think @jcollins1983 was also, so if you're happy with that, then I'll probably head in that direction when writing the migration guide.

test/core/test_switching.py
clint-lawrence: Yep, more tests are on the todo list. See the top comment of the PR. Feel free to edit that comment and add any specific tasks you think off. Or feel free to write some tests and push ;)

src/fixate/drivers/handlers.py
```diff
@@ -0,0 +1,76 @@
...
+    the switching in a jig.
+    """
+
+    def __init__(
+        self,
+        ftdi_description: str,
+        relay_matrix_count: int,
+        extra_pins: Sequence[Pin] = tuple(),
+    ) -> None:
+        relay_matrix_pin_list = tuple(
```
clint-lawrence, line 48: There is a jig that using the follow pins, which match up with schematic designators: RC1K1-16, RC2K1-16, RA1K1-16, RA2K1-16, RA3K1-16, RP1K1-16, RP2K1-16, RH1K1-16

So I'm wondering if we might need to make this a bit more flexible. Although, even as is you could pass in 0 for `relay_matrix_count` and build a pin list anyway you want to pass into `extra_pins`

Perhaps all we need to do is make `_pins_for_one_relay_matrix` a public function and change it to accept a prefix rather than an integer for the designator.

src/fixate/core/switching.py
```diff
@@ -0,0 +1,743 @@
...
+            # and pass in the virtual_map.add_update. But we'd have to do some
+            # magic in the MuxGroup call to pass add_update to each VirtualMux
+            # constructor, and I was hoping to just use a dataclass...
+            mux._update_pins = self.virtual_map.add_update
+
+    def __setitem__(self, key: Pin, value: bool) -> None:
+        self.virtual_map.update_pin_by_name(key, value)
+
+    def __getitem__(self, item: Pin) -> bool:
+        return self.virtual_map[item]
```
daniel-montanari, line 700: I've always hated these due to the lack of clarity of was actually going on and that it seemed like you could bypass everything handled by the virtualmux. I think this implementation is at least a bit nicer now this are a bit less tangled. Would rather an explicit set/get pin if we really want manual control over pins but this should be pretty rare outside of debugging jigs.

clint-lawrence: It is 100% for debugging. Perhaps we should attempt to document than.

Perhaps we should remove these from `JigDriver` and only have them on `jig.virtual_map`. I don't have strong opinions on if we keep the `__getitem__` and `__setitem__` methods, but I think the general functionality is important to keep.

script.py
jcollins1983: Yep, I was in favour of Option 1.

src/fixate/core/switching.py
jcollins1983: I think we should definitely make it clear that these should only be used for debugging.
Because it wouldn't be normal to see something like `jig["pin_name"] = True` I'm not against having the `__getitem__` and `__setitem__` methods.

Maybe prefix methods that are meant only for debugging with debug_ ? Just a thought, but at least it would be very clear that it's for debug only.

src/fixate/drivers/handlers.py
jcollins1983: I like the suggestion to make `_pins_for_one_relay_matrix` public and accepting a prefix. I'm assuming the RA, RC etc have meaning for the specific jig, and there may be other instances where this sort of thing would help identify what matrix is what and I like to have jig schematic and script designations for relays in parity.

src/fixate/core/switching.py
clint-lawrence: I'm leaning toward removing the `__setitem__` and `__getitem__` and adding `jig.debug_set_pin(pin, value)`, `jig.debug_set_pins(on: Collection[Pin], off: Collection[Pin])`

src/fixate/drivers/handlers.py
clint-lawrence: Yeah, those different designators indicate schematic sheet and different PCBs controlled from the one ftdi.

I'll collect a few example and see if I can come up with a simple enough API that fit our current diversity of usage

src/fixate/core/switching.py
```diff
@@ -0,0 +1,743 @@
...
+
+    A virtual switch is a multiplexer with a single pin. The multiplex
+    function can accept either boolean values or the strings 'TRUE' and
+    'FALSE'. The virtual address used to switch can be defined as a list
+    with the single element (as if it were a multiplexer) or by using the
+    shorthand which is to define the pin_name attribute as a string.
+    """
+
+    pin_name: Pin = ""
+    map_tree = ("FALSE", "TRUE")
```
clint-lawrence, line 449: @jcollins1983 @daniel-montanari I'm think of changing this. Most scripts that use a `VirtualSwitch` pass in a bool literal. It seems silly have signal names that are exactly the same. I'm inclined to either do away with the `"TRUE"` and `"FALSE"`. Or turn them into `"On"` and `"Off"`.

Any strong thoughts?

(if it's not obvious, I'm at the point now where there is going to be enough work to migrate existing scripts that we might as well clean up any cruft we can in the process)

clint-lawrence: I've made this change, but happy to roll it back if there is consensus.

daniel-montanari: Not super fussed which way we do things. I just don't like having two ways of doing something.
ELV scripts are the main ones still using "TRUE" and "FALSE"

script.py
daniel-montanari: By defined globally I meant that this doesn't seem to work if you nest it inside a function or class.
It seemed fairly hacky but thought it was interesting in how much you can get away with when forward annotating things and the order that references get resolved.
I'm good with option 1

src/fixate/core/switching.py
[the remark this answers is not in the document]
daniel-montanari: I would think of resetting the jig more like resetting the muxes (which turns all the pins off). Does this mean it turns the pins off but leaves the muxes in their current state? What happens the next time I set the state of a mux and update the pins?

I'm imagining being a grad, resetting the jig and then probing away at stuff and turning one of the muxes back on to check a connection. I'd want the actual behaviour to match the expected behaviour.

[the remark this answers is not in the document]
clint-lawrence: > What happens the next time I set the state of a mux and update the pins?
If you update a mux, it will write update the value of all of its pins (but will not change pins from any other mux)

I think I'm coming back to what I wrote a few comments up
```python
jig.reset()  # reset all muxes to signal "" (via the mux group reset method)

# then if a script really needs it
jig.virtual_map.reset()  # would reset all pins, without changing mux states.
```

> I'd want the actual behaviour to match the expected behaviour.
I just want everyone to agree on the expected behaviour :)

script.py
clint-lawrence: Option 1 it is!

src/fixate/core/switching.py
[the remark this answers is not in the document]
jcollins1983: I'm happy to adopt that approach.

src/fixate/core/switching.py
jcollins1983: I support that

src/fixate/core/switching.py
jcollins1983: I'm happy to keep the change. As far as "switches" go, most people probably think of them in terms of on and off not true and false, unless they're a logic nut ;p

src/fixate/drivers/handlers.py
clint-lawrence: I've changed this around a bit. I've made the handler so that you have to pass in an pin list. Then I've created to helper functions that make it easy to build pin lists similar to the ones I could find in existing scripts.

One potential sticking point - I've not provided the options to use underscores. For example you can only do pins like `RM3K1` not `RM3_K1`. I'm keen to avoid trying to (poorly) cover every possible option and it is a pretty easy search/replace when updating scripts.

src/fixate/drivers/handlers.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    def __init__(
+        self,
+        ftdi_description: str,
+        pins: Sequence[Pin] = tuple(),
+    ) -> None:
+
+        self.pin_list = tuple(pins)
+        # call the base class super _after_ we create the pin list
```
jcollins1983, line 29: This comment is probably not really relevant now that the pin generation stuff isn't here now.

src/fixate/drivers/handlers.py
jcollins1983: Where are the helper functions located? Am I blind?

jcollins1983: Never mind, found them...

jcollins1983: Also, whilst I prefer to have the underscores separator, I think I'm willing to accept its banishment to avoid something like this `f"{prefix}{group_designator}K{relay}" if "_" not in prefix else f"{prefix.strip('_')}{group_designator}_K{relay}" for relay in range(1, pin_count + 1)`
although, I do think that this would cover 95% of the way RMs have been defined to date.

clint-lawrence:
```python
def generate_pin_group(
    group_designator: int, *, pin_count: int = 16, prefix: str = "", sep: str = ""
) -> tuple[Pin, ...]:
    return tuple(
        f"{prefix}{group_designator}{sep}K{relay}" for relay in range(1, pin_count + 1)
    )

generate_pin_group(3, prefix="RM", sep="_")

('RM3_K1',
 'RM3_K2',
 'RM3_K3',
 'RM3_K4',
 'RM3_K5',
 'RM3_K6',
 'RM3_K7',
 'RM3_K8',
 'RM3_K9',
 'RM3_K10',
 'RM3_K11',
 'RM3_K12',
 'RM3_K13',
 'RM3_K14',
 'RM3_K15',
 'RM3_K16')
```

Maybe it's not too bad...

jcollins1983: That's waaaay better than my Frankenstein's monster... 🤦‍♂️

src/fixate/drivers/handlers.py
clint-lawrence: I've modified the comment a little, but it is still relevant that pin_list is assigned before calling super.

But now I'm thinking I should also change the base class so you always pass the pin list into the constructor, instead of the "subclass + declarative pin list" that we've had in the past.

src/fixate/drivers/handlers.py
clint-lawrence: I've relented, but I reserve the right to rant about how we use this inconsistently in the future... https://github.com/PyFixate/Fixate/pull/184/commits/16b81e3acf806f1cd8624f1cdb37e1043218bcb1

src/fixate/core/switching.py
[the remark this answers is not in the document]
clint-lawrence: Done

src/fixate/core/switching.py
```diff
@@ -0,0 +1,792 @@
...
+
+class JigDriver(Generic[JigSpecificMuxGroup]):
+    """
+    Combine multiple VirtualMux's and multiple AddressHandler's.
+
+    The jig driver joins muxes to handlers by matching up pin definitions.
+    """
+
+    def __init__(
+        self,
```
daniel-montanari, line 682: Do we plan on adding a runtime check to make sure all the pins in the muxgroups are implemented in the handlers?

jcollins1983: Could add an assert in the loop that registers `add_update` with each mux.
`assert set(mux.pin_list) <= self.virtual_map._all_pins, "mux pins not in map"`

jcollins1983: Or raise an exception?

clint-lawrence: There is a runtime check that the pin from a virtual mux exists when it gets dispatched to the address handler https://github.com/PyFixate/Fixate/pull/184/files/f464c4e3c5841bb9ff98fae8ed6f937e31f4cec6#diff-9446774e4fed19c7eec4a6f2aec5f67a9ab22433bdbbea3ae51f9359bfa3ee79R608

> For the converse where there are more pins in the handlers than there are in muxes, does it really matter if the pins supplied are sufficient?

I think there is good reason to allow this. It is reasonable to imagine defining "spare" pins. You essentially have to in order for the address handler definition to work.

However I have been thinking of a linting/reporting tool. This could be run manually or part of the script zip process and could error on pins in virtual muxs missing from handlers, warn on unused address handler pins etc

jcollins1983: Was that link above meant to land on this bit of code?
image

jcollins1983: Also, agree RE spare pins.

clint-lawrence: > Was that link above meant to land on this bit of code?
Yep

test/core/test_switching.py
```diff
@@ -0,0 +1,574 @@
...
+def test_generate_pin_group():
+    assert list(generate_pin_group(3)) == (
+        "3K1 3K2 3K3 3K4 3K5 3K6 3K7 3K8 3K9 3K10 3K11 3K12 3K13 3K14 3K15 3K16".split()
+    )
+    assert list(generate_pin_group(1, prefix="RM")) == (
+        "RM1K1 RM1K2 RM1K3 RM1K4 RM1K5 RM1K6 RM1K7 RM1K8 RM1K9 RM1K10 RM1K11 RM1K12 RM1K13 RM1K14 RM1K15 RM1K16".split()
+    )
+    assert list(generate_pin_group(10, pin_count=8, prefix="U")) == (
+        "U10K1 U10K2 U10K3 U10K4 U10K5 U10K6 U10K7 U10K8".split()
+    )
```
jcollins1983, line 64: Each parameter is explicitly tested except for `sep`. Could add this for completeness.
```
assert list(generate_pin_group(10, pin_count=8, prefix="RM", sep="_")) == (
    "RM10_K1 RM10_K2 RM10_K3 RM10_K4 RM10_K5 RM10_K6 RM10_K7 RM10_K8".split()
)
```

jcollins1983: never mind, just found an instance below where sep is included.

clint-lawrence:
O.K. this is no longer draft...

I'm still not 100% on the address handler stuff. But my concerns are really how that ties with #182, so I think I need to get this merged and then see where that PR ends up. If we end up reworking the address handlers in that PR that is O.K.

If also created a handful of smaller issues. Once this PR is reviewed and merged we can decide if we want to do any of that work before making a new release.

src/fixate/core/switching.py
clint-lawrence: `AddressHandler` has moved on a little since this and makes a bit more sense to keep as a subclass rather than switch to a protocol. Also, I suspect we will very rarely need to actually create new sub classes since we will almost always use `fixate.drivers.handlers.FTDIAddressHandler` now.

jcollins1983 (approved):
I'm happy with where things are on this.

mypy.ini
```diff
@@ -66,7 +66,18 @@ exclude = (?x)
             )
         )
     )
+warn_unused_configs = True
+warn_redundant_casts = True
 
+[mypy-fixate.core.switching]
```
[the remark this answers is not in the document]
clint-lawrence: Good catch

script.py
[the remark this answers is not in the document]
clint-lawrence: I'll move it to the examples folder

src/fixate/__init__.py
```diff
@@ -1 +1,33 @@
+# from x import y as y tell linters that we intend
+# to export the symbol as part of our public interface
+# https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface
+
+# Originally, test scripts had to reach into the internal
+# packages and modules for imports. However, we will start
+# to move the API intended for use in test scripts into the
+# top level package namespace and try to be clearer about what
+# is public vs private.
+from fixate.core._switching import (  # noqa
```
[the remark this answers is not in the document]
clint-lawrence: image
Pycharm was giving me a warning.

That is a weird warning really. What point is there having a "private" module if you can't import it at least within the same package!? I'll see if I can switch of that diagnostic and then remove the `noqa` marker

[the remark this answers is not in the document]
clint-lawrence: I've moved `_switching.py` out of the `fixate.core` package, which addresses this warning.

daniel-montanari (approved):

Issue #187, opened 2024-06-06
Alternate concept for VirtualMux


clint-lawrence:
This came about while thinking of other ways to create a VirtualMux that will be type hint and IDE friendly. Originally a comment on #184 but I think it is better to have an issue separate from that so it doesn't get lost.

Each signal would be a method on the resulting VirtualMux subclass. This would be a big change for existing scripts. But once the jig refactor PR lands, we could introduce something like this and scripts could choose which style to use. For new scripts, the IDE autocomplete benefits of this approach would be nice to have. (There might be a way to type-hint the current VirtualMux, but I haven't figured it out)

```python
# existing
class Mux(VirtualMux):
    pin_list = ("x0", "x1", "x2")
    map_list = (
        ("sig1", "x0"),
        ("sig2", "x1"),
        ("sig3", "x2", "x0"),
    )

#new
@dataclass
class MuxNew(VirtualMux):
    sig1: Sig = signal("x0")
    sig2: Sig = signal("x1")
    sig3: Sig = signal("x2", "x0")

dm.jig.mux.Mux("sig2")
# becomes
dm.jig.mux.NewMux.sig2()

# For tree defined multiplexers
#                                   __________
# a0-------------------------------|          |
#                       ________   |          |
# a1_b0----------------|        |--|          |
# a1_b1----------------| Mux B  |  |          |
#              (None)--|  4:1   |  |          |
# a1_b3----------------| x3  x2 |  |          |
#                      |________|  |          |
#                                  |  Mux A   |
# a2-------------------------------|   4:1    |
#                       ________   |          |
# a3_c0----------------|        |--|          |
# a3_c1----------------| Mux C  |  |          |
# a3_c2----------------|  4:1   |  |          |
#              (None)--| x5  x4 |  |          |
#                      |________|  |          |
#                                  |__x1__x0__|
muxA = MuxTree("x0", "x1")
muxB = MuxTree("x2", "x3", nest=muxA)
muxC = MuxTree("x4", "x5", nest=muxA)

@dataclass
class Mux(VirtualMux):
    a0: Sig = muxA()
    a1_b0: Sig = muxB()
    a1_b1: Sig = muxB()
    muxB()  # "throw away" one entry
    a1_b3: Sig = muxB()
    a2: Sig = muxA()
    a3_c0: Sig = muxC()
    a3_c1: Sig = muxC()
    a3_c2: Sig = muxC()
```

[...]

PR #188, opened 2024-06-18
Vmux signal types

daniel-montanari:
Builds on top of  https://github.com/PyFixate/Fixate/pull/184 to address https://github.com/PyFixate/Fixate/issues/187

Rough draft, there are some nuances between 3.8 and 3.10+ due to reworks of the typing library although none of them are gamechangers.

Can keep adding to this if people like this idea.


clint-lawrence:
Obviously we need to get #184 landed, but if this works, we should get this merged too! Then the churn on existing scripts only has to happen once.

For testing, it is hopefully pretty easy to setup some definitions that match the existing "old" style definitions and then compare the resulting `_signal_map`.

src/fixate/_switching.py
```diff
@@ -87,14 +92,26 @@ def __or__(self, other: PinUpdate) -> PinUpdate:
...
-class VirtualMux:
+class VirtualMux(Generic[S]):
     pin_list: PinList = ()
     clearing_time: float = 0.0
 
+    def __class_getitem__(cls, *args, **kwargs):
+        # without calling getitem the class doesn't work as a generic
+        getitm =  super().__class_getitem__(*args, **kwargs) # normally returns a generic
+        # create a proxy to force the __init_subclass__ hook
+        class Hack(getitm):
```
clint-lawrence, line 105: Perhaps rename the `Hack` to `SpecialisedVirtualMux`? Something that reflects its role.

Also need to use `functools.wraps` or whatever secret sauce is required to make errors nice

src/fixate/examples/jig_driver.py
```diff
@@ -58,3 +58,34 @@ class JigMuxGroup(MuxGroup):
...
+except ImportError:
+    # 3.8
+    from typing_extensions import Annotated
+
+from typing import Literal, Union
+
+# maybe we can create aliases to make it easier to understand how to create a MuxDef
+SignalName = Literal
+Signal = Annotated
+MuxDef = Union
```
clint-lawrence, line 73: I think it is probably worth including these in the module and we would document the usage of the class using them.

If nothing else, it will make grepping for places where they are used easier. However, There is a class with the name `Signal`. Perhaps:

```python
SignalName = Literal
SignalDef = Annotated
MuxDef = Union

MuxOneSigDef = MuxDef[
        SignalDef[SignalName["sig_a1"], "a0", "a2" ], 
        SignalDef[SignalName["sig_a2"], "a1"]
    ]
```

clint-lawrence (commented):
I'm pretty wary of the details of `__class_getitem__` breaking. But this is all pretty non-invasive if we don't actually use it in scripts. So I think it is low risk to just go ahead with this.

Big thing we need is some tests that show a virtual mux defined this way ends up with the same `_signal_map` as the original method.

jcollins1983:
We would need to revisit the definition of `name`
image

This is a bit hard to process when all we want is the name of the multiplexer
image

We could do this, which covers for both when the subclass has been annotated and not.
image

For the annotated RelayMatrixMux results in
image

And for the non-annotated MuxA
image

clint-lawrence:
> We could do this, which covers for both when the subclass has been annotated and not. Don't need the `str` cast around `self.__class__.__name__`

Pretty sure that big long class name comes from the call to `new_class` [here](https://github.com/PyFixate/Fixate/pull/188/files#diff-7e7f44c4880743e64f602ecf9d1a32a3e226e3a8f3176fc366beab5b835bd110R123), so we can at least control what it is at the time it's created. I'm not entirely sure what is a good option though...

Now that @daniel-montanari has done the hard work to make both the sub-class and the type alias version work, we should just decide that you need to subclass and be done with it. Then all the complexity goes away at the expense of a slightly more verbose definition.

src/fixate/_switching.py
```diff
@@ -43,15 +43,20 @@
...
 from dataclasses import dataclass
 from functools import reduce
 from operator import or_
 
 Signal = str
+EmptySignal = Literal[""]
 Pin = str
 PinList = Sequence[Pin]
 PinSet = FrozenSet[Pin]
+# do we bother to add EmptySignal here?
```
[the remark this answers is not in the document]
daniel-montanari: I might have written this in the context of Signals being a generic rather than just any string.

src/fixate/examples/jig_driver.py
daniel-montanari: Grepping would be a good use to bother to do this, however extending the behaviour of this classes is basically prevented by a bunch of runtime bits.
Since these are types and not objects it makes it really hard to change the behaviour e.g debug messages will still print the original typing names.

daniel-montanari:
> We would need to revisit the definition of `name` image
> 
> This is a bit hard to process when all we want is the name of the multiplexer image
> 
> We could do this, which covers for both when the subclass has been annotated and not. Don't need the `str` cast around `self.__class__.__name__` image
> 
> For the annotated RelayMatrixMux results in image
> 
> And for the non-annotated MuxA image

I've changed the behaviour of this to digest all the type info upfront which then results in a class the same as if you had defined the signals manually.

I still chucked the signal definition to give a differentiation of which mux it should be, but that still results in really long error messages, more to come...

src/fixate/_switching.py
```diff
@@ -87,14 +93,73 @@ def __or__(self, other: PinUpdate) -> PinUpdate:
...
+        # __orig_bases__ is an attribute set by the metaclass that is used to store the type information provided by __class_getitem__
+        # expected class creation:
+        # VirtualMux[type] -> VirtualMux.__class_getitem__(type) -> VirtualMux.__init__() (with type info visible on __orig_bases__)
+        # what actually happens
+        # VirtualMux[type] -> VirtualMux.__class_getitem__(type) -> _GenericAlias(VirtualMux).__call__() -> VirtualMux.__init__()
+        # the _GenericAlias middle class means we lose the type information provided through the __class_getitem__ call
+
+        # the attribute __orig_class__ can be used to store the original class that was created with __class_getitem__
+        # HOWEVER this relies on the __init__ of VirtualMux succeeding - we end up in a circular dependency of needing
+        # the init to succeed to set the attribute, but needing the attribute to be set to succeed in the init (due to our implementation)
```
clint-lawrence, line 118: General code organisation point. Can you please move this down to the bottom of the class definition where the other internal details are?

Also, please wrap the comments to ~100 columns.

src/fixate/_switching.py
```diff
@@ -104,16 +169,16 @@ def __init__(self, update_pins: Optional[PinUpdateCallback] = None):
         else:
             self._update_pins = update_pins
 
+        self._pin_set = frozenset(self.pin_list)
```
clint-lawrence, line 172: Not sure why this has moved around, but not the is in the wrong place. It should probably just move back where it was.

daniel-montanari: I moved this in the previous commit due to playing around with some things. It should be back to where it was originally now.

src/fixate/_switching.py
```diff
@@ -441,6 +456,83 @@ def _default_update_pins(
         """
         print(pin_updates, trigger_update)
 
+    # This looks bad, maybe we should not type hint this
+    def __class_getitem__(cls, arg: Union[TypeVar, Any]) -> type[VirtualMux[S]]:
```
clint-lawrence, line 460: I think it should be `arg: Union[TypeVar, Type]`

src/fixate/_switching.py
daniel-montanari: We can revisit this idea when we have a better way of getting around GenericAlias stripping all annotations at runtime.

src/fixate/examples/jig_driver.py
daniel-montanari: I thought this is what you meant earlier today with aliases.
Mypy doesn't like assigning things like `Literal` to a variable. Assigning the mux definition to a variable is ok since that's a properly defined type.

src/fixate/_switching.py
```diff
@@ -87,14 +93,73 @@ def __or__(self, other: PinUpdate) -> PinUpdate:
...
+        # HOWEVER this relies on the __init__ of VirtualMux succeeding - we end up in a circular dependency of needing
+        # the init to succeed to set the attribute, but needing the attribute to be set to succeed in the init (due to our implementation)
+
+        # there are two options
+        # 1. create a proxy class and wrap it to look like a normal VirtualMux
+        # 2. dynamically create the pin_list map_list here
+
+        proxy = new_class(f"{cls}[{arg}]", bases=(cls,))
+        # two cases: either we are a passing in the generic typevar, or we are passing in the mux definition
+        if type(arg) != TypeVar:
```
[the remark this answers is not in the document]
daniel-montanari: Changed it to do this, but idea is dead now 😢

src/fixate/_switching.py
daniel-montanari: I figured out how to get a ruler in vscode so will happen way less often now if I forget to run my formatter.

src/fixate/_switching.py
daniel-montanari: Technically this is correct and counterintuitively here I think it makes more sense that the type hint here is for what happens at runtime so that you can have useful introspection inside the function.
It seems like the inheritance from Generic is what sets the type hint on this call for tools like mypy

daniel-montanari:
Given up on support for 3.8.
Push this back until after we switch to 3.12

daniel-montanari:
I hope the deploy failing is because my code breaks on 3.8 and not actually because it couldn't install Node.js

jcollins1983:
This comment is far enough away from the other one that gives it context that I think it warrants a rewording.
image

Maybe `# RelayMatrixMux is an example of a Generic class being reused.`

src/fixate/_switching.py
```diff
@@ -43,17 +43,33 @@
     Dict,
     FrozenSet,
     Iterable,
+    Literal,
+    get_origin,
+    get_args,
+    Any,
```
jcollins1983, line 49: Looks like Any is not used any more

src/fixate/_switching.py
```diff
@@ -43,17 +43,33 @@
...
 )
 from dataclasses import dataclass
 from functools import reduce
 from operator import or_
+from types import new_class
```
jcollins1983, line 54: Can delete the import of `new_class` now?

src/fixate/_switching.py
```diff
@@ -43,17 +43,33 @@
...
+
+import sys
```
jcollins1983, line 56: Assuming this conditional import won't be necessary if we're heading straight for 3.12?

clint-lawrence:
> I hope the deploy failing is because my code breaks on 3.8 and not actually because it couldn't install Node.js

You can drop the other python versions from CI test matrix on this branch if you want.

src/fixate/_switching.py
```diff
@@ -87,15 +103,23 @@ def __or__(self, other: PinUpdate) -> PinUpdate:
...
+    map_list: Optional[Sequence[Sequence[str]]] = None
     pin_list: PinList = ()
     clearing_time: float = 0.0
 
     ###########################################################################
     # These methods are the public API for the class
-
     def __init__(self, update_pins: Optional[PinUpdateCallback] = None):
+        # digest all the typing information if there is any to sete pin_list and map_list
+        self._digest_type_hints()
```
jcollins1983, line 121: I think Clint mentioned this before, and I think you said it should be back where it was, but the comment still doesn't look like it's in the right place to me. Shouldn't it be above `self._pin_set`?

src/fixate/_switching.py
```diff
@@ -441,6 +469,50 @@ def _default_update_pins(
...
+        elif get_origin(muxdef) == Annotated:
+            # Union FORCES you to have two or more types, so this handles the case of only one pin
+            signals = (muxdef,)
+        else:
+            raise TypeError("Signal definition must be Union or Annotated")
+
+        map_list: list[tuple[str]] = []
+        pin_list: list[str] = []
+        for s in signals:
+            # not working, get_origin(Annotated) returns the wrapped type in 3.8
```
jcollins1983, line 500: Is this comment still relevant?

daniel-montanari:
> > I hope the deploy failing is because my code breaks on 3.8 and not actually because it couldn't install Node.js
> 
> You can drop the other python versions from CI test matrix on this branch if you want.

Yep was code not working due to python versions not anything to do with the build process.

clint-lawrence:
How's the migration to 3.13 going, because this would nice to return to :)

clint-lawrence:
Awww yeah, good to see this get resurrected :)

daniel-montanari:
I've noticed an annoying edge case. Defining the definition globally has a different inferred type compared to locally in a function.

I think the reason for this is due to how typealiastypes get created
image
Here's the same thing. X is not a variable, it is a typealiastype. x is the class int, you can't use it in generics, while you can use X.
image


image

image

clint-lawrence:
It might work to do `type muxdef = Annotated...` inside the function body, if you really want to reduce the scope

PR #194, opened 2024-07-19
Docs: add API reference based on docstrings

josh-marshall-amp:
I saw this [internal doc](https://ampcontrol.atlassian.net/wiki/spaces/CPEK/pages/445382657/Using+Fixate+Drivers+in+Python) pop up, and went ... hey, isn't that what good API docs are for? And given this is on Github, pretty sure we can automate it, and then it'll be kept up to date, included in the release page and if docstring-based will also be accessible in IDEs.

So this PR:
- [x]  uses `sphinx-apidoc` to generate the API docs
- [x] integrates them into the existing docs
- [x] tweaks some of the docstrings to demo the potential (using [Google style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html))

And in order to make it actually useful, either as part of this or as a followup, should:

- [ ] tidy up the way the docs actually present (on the .rst side)
- [ ] fix, then add a stack more docstrings, at least enough to cover the basics and provide an example
- [ ] use a workflow to autogen the docs on a PR to main
- [ ] include docs in release package
- [ ] nab the allocated fixate github.io subdomain and push it there

## Sample

DMM_package

DMM_helper

## References
- [sphinx-autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html)
- [sphinx-apidoc](https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html)
- [idiot's guide (ie me)](https://samnicholls.net/2016/06/15/how-to-sphinx-readthedocs/)


docs/conf.py
```diff
@@ -15,20 +15,22 @@
 import os
 import sys
 
+import fixate
```
jcollins1983, line 18: Pretty sure this import needs to be after the `sys.path.append` otherwise it won't be able to find the fixate module.

jcollins1983 (commented):
I'm in favour of improving our documentation. I'd also be in favour of automating the build of the documentation as part of the PR process.

josh-marshall-amp:
@josh-marshall-amp I deliberately overrode pre-commit for my commits, since there is a stack of stuff to fix overall and I wanted to get the concept up for feedback first... no point spending that time in case it was going counter to other folks' views on the direction for fixate. So I've marked them resolved without fixing first... but would do so before merge.

josh-marshall-amp:
@jcollins1983 I also like to keep 'content' vs 'process' PRs separate... so my intention with the changes to the docstrings themselves were just as an example, and I can imagine those could be updated by those who know Fixate or are using it more than I. We should discuss whether Google- or numpy-style docstrings are a preference, as they seem to be online (much more readable and "pythonic")

docs/conf.py
josh-marshall-amp: 🤦‍♂️ this is why one _automates_ things, right? 😉

docs/quickstart.rst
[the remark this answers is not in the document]
clint-lawrence: I'm vaguely in favour of keeping `TEST_SEQUENCE`. I think it's nice to have a simpler interface to testing and experimenting. If it's more confusing than helpful, we can can it though

josh-marshall-amp:
- Added the docs build to `tox.ini`. (I'm still running it with a manual sphinx-build invocation so I get the right filenames to fix, but the automation works)
- added the special-casing so it doesn't spew errors on unimportable packages (PyQt, PyDAQ) and pulled the ftdi low-lever driver into its own `_ftdi.py` so that can be mocked as well
- a few more simple fixups to docstrings

There's still quite a number of warnings on docstrings, so these will need to be cleaned up at some stage.

Still need the discussion on docstring style... but it does look like a mix of rst :param name: and the google style is acceptable to Sphinx, can migrate slowly.

Haven't added the docs built to the github workflow, but that's probably worthwhile even if we're not using them just yet. I don't want to fail builds on doc warnings yet though.

tox.ini
```diff
@@ -34,3 +34,12 @@ deps = mypy==1.10.0
...
 # separate cache to stop weirdness around sharing cache with other instances of mypy
 commands = mypy --cache-dir="{envdir}/mypy_cache" --config-file=mypy.ini
+
+[testenv:docs]
+description = invoke sphinx-build to build the HTML docs
+basepython = python3
+deps = sphinx >= 7.4, < 8
+# once docstrings are all up-to-date, add the -W flag to turn warnings into errors
+commands =
+    sphinx-build -d "{toxworkdir}/docs_doctree" docs "{toxworkdir}/docs_out" --color -bhtml {posargs}
```
clint-lawrence, line 44: You can also set `skip_install = True`, then run tox on the src tree directly. I don't think there is any benefit in running sphinx against the installed source
https://tox.wiki/en/latest/config.html#skip_install

josh-marshall-amp: Adding `skip_install = True` to `tox.ini`/`[testenv:docs]` means the whole docs build fails because things like `platformdirs` function calls aren't mocked correctly, even after adding practically all the requirements to `conf.py`:`autodoc_mock_imports`.

The ones I've mocked (ftdi, PyDAQmx, PyQt) are because they are platform-dependent / do something fancy at import time.

`CONFIG_DIRECTORY = Path(platformdirs.site_config_dir("Fixate", False))` would be a PITA to mock

clint-lawrence: Ack

clint-lawrence: Can we pin a specific base python here? The reason I went down the path suggested above, was because this failed for me originally. Now, is in some way my own fault because I have python 3.13 beta installed, and tox decided to use that.

If we can pin basepython to 3.12 it would be nice. Especially when we start to build locally and in CI it's nice to have stuff pinned.

clint-lawrence (commented):
I'm generally in favour of doing running this as is. Not sure how to test it against read the docs though (I think when we merge into master it will kick off a build, but not sure)

Plenty of details to tweak and improve, but I'm happy to merge and iterate, rather than try to perfect everything here. Happy to be guided by you and what you think are the biggest gaps with where this is right now

clint-lawrence:
I've enabled PR preview builds on read the docs. Hopefully that works on the next push, but there might be other config needed. I'm not 100% sure.

https://docs.readthedocs.io/en/stable/pull-requests.html

josh-marshall-amp:
I've pinned the docs build to 3.12 as requested. Confess I don't really quite get basepython vs the evnlist matrix... and how all that interacts with the build given we're still using the 3.8 build?

clint-lawrence (approved):
@jcollins1983 & @josh-marshall-amp the read the docs builds are now working, including previews on PRs.

e.g.
https://fixate--194.org.readthedocs.build/en/194/fixate/fixate.drivers.dmm.html

Plenty of work todo, mainly removing sections that aren't really a part of the public api. I think this is a great step in the right direction.

jcollins1983:
> Still need the discussion on docstring style... but it does look like a mix of rst :param name: and the google style is acceptable to Sphinx, can migrate slowly.

I don't really have a strong preference, but I think I lean towards Google style, it looks a bit easier to read IMO.

docs/conf.py
```diff
@@ -15,20 +15,20 @@
...
-project = 'Fixate'
-copyright = '2018, Clint Lawrence, Ryan Parry-Jones'
-author = 'Clint Lawrence'
+project = "Fixate"
+copyright = "2024, Clint Lawrence, Ryan Parry-Jones"
+author = "Clint Lawrence"
 
 # The short X.Y version
-version = ''
+version = "" #fixate.__version__
```
jcollins1983, line 29: Do we need to specifiy `version` and `release` at all if the default is `''`?

jcollins1983: Also, I'm bumbling through the sphink documentation so try and understand all this better, so probably a lot of detail/understanding that I'm missing at this stage.

docs/conf.py
```diff
@@ -15,20 +15,20 @@
 import os
 import sys
 
-sys.path.append(
-    os.path.join(os.path.dirname(__file__), '..', '..', 'src')
-)
+# sys.path.append(os.path.join(os.path.dirname(__file__), "..", "src"))
+#
+# import fixate
```
jcollins1983, line 20: So, because the tox setup somehow magically knows how to install fixate we don't need import (satisfying option 1 [here](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html)) . It doesn't work though if one was to use the Makefile, assuming that the primary way to build the documentation will be via tox, do we need to keep the Makefile? Or the make.bat file?

Is there any reason to leave the unused import here?

clint-lawrence: The answer to both you questions is that I was lazily changing things trying to fix a few unrelated issues.

But I think I've made it all better now... watch for an incoming PR

docs/conf.py
clint-lawrence: I've changed this back to it uses `fixate.__version__` to set the `version` and `release` attributes.

docs/conf.py
josh-marshall-amp: @jcollins1983 I've just been doing a `sphinx-build . _build` anyway. I think I noted it elsewhere, but the disadvantage I've found with `tox -e docs` is that the warnings end up being:

```
/Users/josh/Ampcontrol/repos/pyfixate/.tox/docs/lib/python3.12/site-packages/fixate/drivers/funcgen/keysight_33500b.py:docstring of fixate.drivers.funcgen.keysight_33500b.Keysight33500B.get_identity:3: WARNING: Definition list ends without a blank line; unexpected unindent.
```
instead of
```
/Users/josh/Ampcontrol/repos/pyfixate/src/fixate/drivers/funcgen/keysight_33500b.py:docstring of fixate.drivers.funcgen.keysight_33500b.Keysight33500B.get_identity:3: WARNING: Definition list ends without a blank line; unexpected unindent.
```
where in the tox version it's too tempting to click on the source file to change it... but that's the copy tox has made to build.

@clint-lawrence, if tox doesn't install the source or whatever magic you've done, presumably this problem goes away too?

josh-marshall-amp: To answer my own question... yes that PR fixed it...

```
/Users/josh/Ampcontrol/repos/pyfixate/docs/../src/fixate/drivers/funcgen/keysight_33500b.py:docstring of fixate.drivers.funcgen.keysight_33500b.Keysight33500B.get_identity:3: WARNING: Definition list ends without a blank line; unexpected unindent.
```
works!

clint-lawrence: The `import fixate` is to be able to set the `version` and `release` attributes. They were temporarily commented out while I was trying to fix a dependency problem.

Manipulating the python path allows building the docs over the source tree, instead of the installed python files in a virtual env.

josh-marshall-amp:
I added a docs readme. Besides that, anything else for this PR? 

I suggest we merge this, and then the next discussion can be about the structure and style of the docs themselves.

There was the [discussion re: style for the examples](https://github.com/PyFixate/Fixate/pull/194#discussion_r1685700632) but that can wait too.

clint-lawrence:
> I added a docs readme. Besides that, anything else for this PR?
> 
> I suggest we merge this, and then the next discussion can be about the structure and style of the docs themselves.
> 
> There was the [discussion re: style for the examples](https://github.com/PyFixate/Fixate/pull/194#discussion_r1685700632) but that can wait too.

Yes to all of that from me.

The only unresolved comments are from @jcollins1983 about if we should keep or remove `TEST_SEQUENCE`. I think that is a discussion for elsewhere.

jcollins1983:
> > I added a docs readme. Besides that, anything else for this PR?
> > I suggest we merge this, and then the next discussion can be about the structure and style of the docs themselves.
> > There was the [discussion re: style for the examples](https://github.com/PyFixate/Fixate/pull/194#discussion_r1685700632) but that can wait too.
> 
> Yes to all of that from me.
> 
> The only unresolved comments are from @jcollins1983 about if we should keep or remove `TEST_SEQUENCE`. I think that is a discussion for elsewhere.

Agree that the conversation RE `TEST_SEQUENCE` can happen elsewhere/later.

docs/readme.md
```diff
@@ -0,0 +1,105 @@
...
+   - example_walkthrough (tutorial)
+   - example projects (how-to)
+   - fixate/fixate (reference in-depth API docs, see below)
+   - release-notes (reference)
+
+# Building
+
+- Uses [Sphinx](https://www.sphinx-doc.org/)
+- `tox -e docs` will build, see `tox.ini::[testenv:docs]` section
+    - underneath it, with an appropriate venv: `fixate/docs > sphinx-build . _build`
```
jcollins1983, line 30: This is my last issue. It's not really clear what this line is trying to convey. Is it supposed to indicate a way to build without tox or is it supposed to be indicating what tox is doing under the hood?
If it's the former, I don't think it should be a sub-point of `tox -e docs`. If it's something else entirely, some extra clarity wouldn't go astray for the numpties like me.

josh-marshall-amp: > Is it supposed to indicate a way to build without tox or is it supposed to be indicating what tox is doing under the hood?

Yeah, under the hood. Clarified now.

docs/requirements.txt
josh-marshall-amp, line 1: @clint-lawrence I added numpy and pynput (and some others) to the mocked modules in `docs/conf.py`, so these don't need to be installed into the docs build venv.

Likely I'm nitpicking and it doesn't actually matter once all the caching etc is taken into account and the fact that it's CI and nobody is sword-fighting while they wait for their docs to compile, but there was a significant time + disk space difference for me.

clint-lawrence: Yeah, but they are still in the requirements file, so will still get installed, right? I have zero enthusiasm to prune them from the requirements file...

josh-marshall-amp: yep, I figured.

PR #196, opened 2024-07-22
add changes from black

josh-marshall-amp:
These are the actual changes from adding pre-commit and doing an `--all-files` run.

I spotted this because @jcollins1983 was calling me out for corrections that black should have done and me thinking "but I have pre-commit" running.


josh-marshall-amp:
tl;dr — `ruff`

I also added `ruff-pre-commit` to the hooks, but it made far too many changes across the codebase. And a lot of the samples have unused includes, but it's because they are likely to be used etc.

A thought I had is that I could have it installed locally via pre-commit but _not_ in the repo checks, for now at least.

If you want to try it, I was using: (note version is much updated from our other repos, it was at 0.0.28!) 
```
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.5.4
    hooks:
      - id: ruff
        args: [--fix, --exit-non-zero-on-fix]
```

tox.ini
```diff
@@ -16,7 +16,7 @@ commands =
 [testenv:black]
 basepython = python3
 skip_install = true
-deps = black==22.3.0
+deps = black==24.4.2
 commands = black --check src test scripts {posargs}
```
clint-lawrence, line 20: This PR + adding pre-commit has become a bit muddled.

The only reason these formatting changes have crept in, is because pre-commit config has the latest black. If you change the pre-commit config to use 22.3.0 everything should have checked just fine. So this PR is really "bump black version to the latest and reformat"

I'm keen to have pre-commit as part of the repo, but don't want to have to maintain two sets of config. We could change the black environment in tox to run pre-commit. Then the workflow can stay the same. Alternatively we could change the workflow definition to run pre-commit directly.
https://github.com/PyFixate/Fixate/blob/master/.github/workflows/test.yml#L37

clint-lawrence (changes requested):

josh-marshall-amp:
Closed in lieu of https://github.com/PyFixate/Fixate/pull/199 which doesn't have all the back-and-forth history garbage.

PR #197, opened 2024-07-22
Parameterise sys args

josh-marshall-amp:
Proof of concept... let test scripts act as their own runner.

```python
if __name__ == "__main__":
    import fixate.__main__

    argv = ["-p", __file__, "--serial_number", "1234567890"]
    fixate.__main__.run_main_program(__file__, argv)
```

The only significant change was adding `main_args` to `run_main_program(test_script_path=None, main_args=None)`, which has the same default behaviour, but if the args are passed in then it uses those instead of `sys.argv`.

I also added the argument parser build into its own function, and deglobalised `parser`.

Steps were being taken in this direction elsewhere, but it was quick enough to try out and make sure things don't break.

### DIFF
--- a/src/fixate/examples/tiny.py
+++ b/src/fixate/examples/tiny.py
@@ -41,3 +41,9 @@ class ParameterisedTest(TestClass):
 
 
 TEST_SEQUENCE = [SimpleTest(), MyTestList([ParameterisedTest(1), ParameterisedTest(2)])]
+
+if __name__ == "__main__":
+    import fixate.__main__
+
+    argv = ["-p", __file__, "--serial_number", "1234567890"]
+    fixate.__main__.run_main_program(__file__, argv)
--- a/src/fixate/examples/tiny.py
+++ b/src/fixate/examples/tiny.py
@@ -43,7 +43,7 @@ class ParameterisedTest(TestClass):
 TEST_SEQUENCE = [SimpleTest(), MyTestList([ParameterisedTest(1), ParameterisedTest(2)])]
 
 if __name__ == "__main__":
-    import fixate.__main__
+    import fixate
 
     argv = ["-p", __file__, "--serial_number", "1234567890"]
-    fixate.__main__.run_main_program(__file__, argv)
+    fixate.run(__file__, argv)
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ venv37
 .vscode/
 .venv
 .vscode/settings.json
+docs/_build/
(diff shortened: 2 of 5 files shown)


src/fixate/examples/tiny.py
```diff
@@ -41,3 +41,9 @@ def test(self):
 
 
 TEST_SEQUENCE = [SimpleTest(), MyTestList([ParameterisedTest(1), ParameterisedTest(2)])]
+
+if __name__ == "__main__":
+    import fixate.__main__
+
+    argv = ["-p", __file__, "--serial_number", "1234567890"]
+    fixate.__main__.run_main_program(__file__, argv)
```
clint-lawrence, line 49: How about having a nicer alias for that function?

e.g. adding the following to the top level `__init__.py`
`from fixate.__main__ import run_main_program as run`

So then the snippit above would become
```python
if __name__ == "__main__":
    import fixate

    argv = ["-p", __file__, "--serial_number", "1234567890"]
    fixate.run(__file__, argv)
```

clint-lawrence (approved):
Looks good to me.

I was thinking of this when doing the other work on the amptest `run` function, but didn't feel like a side quest at the time. We will also need add a cli flag to override the config file default, but that can wait for another day.

src/fixate/examples/tiny.py
josh-marshall-amp: Done. Now the only thing bugging me on that is that with

```
argv = ["-p", __file__, "--serial_number", "1234567890"]
fixate.run(__file__, argv)
```

we are actually supplying the test file twice: as both `test_script_path` arg and as the script itself to run with `-p xyz.py`. It still works if we skip `test_script_path` but _not_ if we don't provide the `-p path`

`__main__.py:283-284,296-298` has:

```python
            if self.test_script_path is None:
                self.test_script_path = self.args.path
            # [… snip …]
            # Load test suite
            test_suite = load_test_suite(
                self.args.path, self.args.zip, self.args.zip_selector
            )
```

where `self.test_script_path` is not used further.  Am I on the right path (pun intended) with thinking that the `load_test_suite()` call should have `self.test_script_path` not `self.args.path` passed?

clint-lawrence: > Am I on the right path (pun intended) with thinking that the load_test_suite() call should have self.test_script_path not self.args.path passed?

Looks like it. But much of that code is a mess that could be simplified. I've made a few passes over it in the past, but never just got stuck in a properly refactored. I'm not super keen to try fix all this right now. At least not without some kind of big picture goal. Which at the moment, is the `atdev`

PR #198, opened 2024-07-22
Fix #167: Check if stdin is a terminal before calling termios on it

### LINKED ISSUE #167: Tests fail on MacOS
# The tests fail on MacOS
(and possibly on other *nix systems)

An exception is raised on access to attributes in termios.
image

When running normally, this code does not raise an exception. Not sure why pytest isn't happy with it... yet.

Doesn't appear to be isolated to pytest though... https://github.com/pfalcon/picotui/issues/71 

Since there's no usage (that we know of) of fixate on *nix systems, it's probably not worth chasing down, but I wanted to note it.

josh-marshall-amp:
On *nix, check if stdin is a terminal before calling termios on it as those operations fail on a pipe, ie when running pytest.

### DIFF
--- a/src/fixate/ui_cmdline/kbhit.py
+++ b/src/fixate/ui_cmdline/kbhit.py
@@ -39,8 +39,7 @@ class KBHit:
         if os.name == "nt":
             pass
 
-        else:
-
+        elif os.isatty(sys.stdin.fileno()):
             # Save the terminal settings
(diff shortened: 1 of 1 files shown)


clint-lawrence (approved):
Looks good to me.

*Ages* ago, I had tried to keep this running on MacOS and Linux, so more than happy for small fixes to get that back.

PR #202, opened 2024-07-23
Fix the __main__ entry point.

clint-lawrence:
Because we were importing `__main__` in `__init__`, we ended up with a warning that `__main__` was already imported when it was executed as main. Moving the previous contents of `__main__.py` to `main.py` so that it can then be import in `__init__.py` and `__main__.py` fixes that.


clint-lawrence:
@josh-marshall-amp the github diff makes it hard to review, but nothing has changed between __main__.py and main.py other than removing the `if __name__ == "__main__"` block from the bottom.

src/fixate/__init__.py
```diff
@@ -30,6 +30,6 @@
     generate_relay_matrix_pin_list as generate_relay_matrix_pin_list,
 )
 
-from fixate.__main__ import run_main_program as run
+from fixate.main import run_main_program as run
```
jcollins1983, line 33: What does this import achieve?

clint-lawrence: It exports the function to be part of the public api and allows creating script to be directly executed. Similar to what I've done in `atdev`, but at the fixate level. Once this is released, I'll use this (and the `--diagnositic-log-dir` argument) to fix the hackiness in `atdev`. Josh added an example also:
https://github.com/PyFixate/Fixate/blob/master/src/fixate/examples/tiny.py#L45C1-L49C31

I'd like to slowly make it clearer what is public api vs private implementation detail.

clint-lawrence:
TIL - you can diff a renamed file across two branches:
```sh
> git diff master:./__main__.py ./main.py
diff --git a/src/fixate/__main__.py b/src/fixate/main.py
index f446c2c..d5db95a 100644
--- a/src/fixate/__main__.py
+++ b/src/fixate/main.py
@@ -424,8 +424,3 @@ def run_main_program(test_script_path=None, main_args=None):
     # Could this be replaced with a simple log_file variable in fixate.config ^ ?
     supervisor = FixateSupervisor(test_script_path, args)
     exit(supervisor.run_fixate())
-
-
-# Setup configuration
-if __name__ == "__main__":
-    run_main_program()

```

src/fixate/__main__.py
josh-marshall-amp, line 1: nitpick history remark: if you move `__main__.py` to `main.py` in one commit, then in a separate commit add the "new" `__main__.py`, the diff will show up more clearly.

josh-marshall-amp (approved):
good stuff

PR #203, opened 2024-07-23
Diagnostic log argument

clint-lawrence:
This allows overriding the default diagnostic log location.

In the atdev tool I'm working on, I'm currently monkey patching the log location. This will allow removing that.


jcollins1983 (approved):
Looks good

PR #206, opened 2024-07-31
Move examples out of the package and update readme

clint-lawrence:
This came up somewhere in the threads discussion the `VirtualMux` related changes.


clint-lawrence:
@daniel-montanari tagged you for require, because I think you suggested this? or at least we discussed it.

daniel-montanari (changes requested):
Looks good, can we remove the examples folder from `mypy.ini` exclusions now please.
[mypy.ini](https://github.com/PyFixate/Fixate/blob/move-examples/mypy.ini)

I believe the rules on the example files at the bottom are also no longer required.

clint-lawrence:
> I believe the rules on the example files at the bottom are also no longer required.

Should be good now.

daniel-montanari (approved):
Looks good

PR #212, opened 2024-08-11
Fxconfig duplicate prevention

jcollins1983:
This PR addresses the TODO `# TODO: Prevent writing duplicate to the config.` and adds some tests.

The `{posargs}` in the `pytest` call in `tox.ini` was committed by accident, it was part of my debugging of testing on mac, but I left it in as I figured it isn't hurting anything and provides some flexibility. Happy to revert if requested though.


clint-lawrence (approved):
I think this is good step forward.

The tests might prove to be a bit brittle, but I think that is a side effect of how the code was originally written. Any future effort to improve the code structure will hopefully benefit from the tests!

Issue #213, opened 2024-08-14
_user_req_input ignores valid falsy values


daniel-montanari:
The ui implementations of requesting user input have a flaw when returning results.
This is due to the validator functions returning a Union of validated input and False . The ui functions then blindly convert the result to a boolean and loop without feedback if a false value is received from the input validator. If a valid input converts to false, then the input is discarded.

I found this happening specifically with `user_input_float` as 0 or 0.0 are valid floats.

In `cmd_line.py`
```python
    for _ in range(attempts):
        # This will change based on the interface
        print("\a")
        ret_val = input(msg)
        if target is None:
            q.put(ret_val)
            return
        ret_val = target(ret_val, **kwargs)
        if ret_val:
            q.put(("Result", ret_val))
            return
```
In `ui_gui_qt.py`
```python
        for _ in range(attempts):
            # This will change based on the interface
            ret_val = self.gui_user_input(msg, None)
            if target is None or ret_val == "ABORT_FORCE":
                q.put(ret_val)
                return
            ret_val = target(ret_val, **kwargs)
            if ret_val:
                q.put(("Result", ret_val))
                return
```

jcollins1983:
This issue was resolved by #214 which has now been merged.

Issue #177, opened 2023-09-20
Do away with the need to use indexing on user response functions


jcollins1983:
This may be more involved than it sounds.

Refer to ui_gui_qt.py
``` 
def _topic_UI_req_choices(self, msg, q, choices, target, attempts=5):
    """
    This can be replaced anywhere in the project that needs to implement the user driver
    Temporarily a simple input function.
    The result needs to be put in the queue with the first part of the tuple as 'Exception' or 'Result' and the
    second part is the exception object or response object
    This needs to be compatible with forced exit. Look to user action for how it handles a forced exit
    ... 
```

The requirement for the item to be put in the queue as a Tuple with `Exception` or `Result` as the first element is the thing that drives the necessity to do things in test scripts like this `relay_open = user_yes_no("Confirm the relay is open. (DMM not buzzing)")[1]`

The same is done in the cmd_line.py, where `_user_choices` has the same docsctring. Understanding why this was thought (or is) necessary and/or whether it actually requires this format would be good. I'd like to dig in because I'm interested but can't right now.

jcollins1983:
`_user_req_choices` in `ui.py` is the mechanism used by `user_yes_no` and `user_retry_abort_fail`. 
`_user_req_choices` looks after the messaging between the actual UI with the topic of `UI_req_choices`.
In `cmd_line.py` the listener is `_user_choices` and in `ui_gui_qt.py` the listener is `_topic_UI_req_choices`.

Both have identical logic, where a loop iterates until the user selects a valid choice. In the gui the user has no option other than to select a valid choice, so the message put on the queue will always be `("Result", ret_val)`.

In the command line, it is possible to enter an invalid choice, however when the number of retries is exceeded (default for `user_yes_no` is 1) the returned value is only the string `"Exception"`.

<img width="549" alt="image" src="https://github.com/user-attachments/assets/747b03f8-7704-4a8b-a932-be3e76941298">

For some reason the UserInputError exception object is not ending up in the queue.
<img width="574" alt="image" src="https://github.com/user-attachments/assets/786014d2-78e0-4cd6-a589-5dffc1c21a34">

And, when called by `user_retry_abort_fail` when the number of attempts are exceeded a different error is encountered.
<img width="564" alt="image" src="https://github.com/user-attachments/assets/89f1aeaf-67a4-44f9-bf45-8e692c10323a">

I'll come back to this soon.

clint-lawrence:
[...]

PR #214, opened 2024-09-01
Revamp UI interactions

jcollins1983:
This PR addresses Issue #177 and Issue #213.

I've done a similar thing to the switching module and put the new _ui.py module at the fixate level and have only exposed the public functions.

To avoid having 2 different sets of functions in `cmd_line.py` and `ui_gui_qt.py` I moved the (duplicated) logic out of the UI files into ui.py and kept the original behaviour, tests were updated to deal with this change.

Also added the ability to have colour on `user_info_important` the line of `!` defaults to red, which will hopefully be a bit more eye catching.

Have tested each of the functions manually. But should try on a real script or 2 before this gets merged if it does.


jcollins1983:
Looks like only the `readthedocs` build is done, no other tests? 🤔
image

jcollins1983:
This'll be why
image

jcollins1983:
> This'll be why image

I'll do another targeted branch to fix this.

clint-lawrence:
I haven't looked at the details here yet. General idea look fine.

This is what I'm thinking - there is probably a fair bit more we can/should do to improve. For example, tighten up typing to be more specific instead of many parameters/return values being strings. I don't want to get this improvement bogged down in trying to fix everything, but I don't want to shoot ourselves in the foot and complicate future improvements.

In any case, I'll get in and review this as is. And hopefully none of that is an issue.

src/fixate/ui_gui_qt/ui_gui_qt.py
```diff
@@ -839,23 +775,32 @@ def _topic_UI_display(self, msg):
...
         """
         :param msg:
         :return:
         """
         if self.closing:
             return
 
+        txt_fill = "!" * wrapper.width
+
+        # Hisotry window
```
daniel-montanari, line 788: how could you

src/fixate/ui_gui_qt/ui_gui_qt.py
```diff
@@ -738,96 +738,32 @@ def _topic_UI_action(self, msg, callback_obj):
...
-         Validation function to check if the user response is valid
-        :param attempts:
-        :return:
-        """
+        self._gui_user_input(msg, ("Continue",))
+
+    def _topic_UI_req_choices(self, msg, q, choices):
         if self.closing:
+            # I don't think the result of this code ever gets used, nothing looking for "ABORT_FORCE"
+            # unpacks from a tuple. This can probably be deleted.
```
daniel-montanari, line 750: 2 things in main look for this, I really don't understand what these are doing though. Doesn't make sense to me that we would inidicate a forced abort during a user input function

src/fixate/ui_gui_qt/ui_gui_qt.py
daniel-montanari, line 1: Do you want to give `reformat_text` the same treatment you did in the cli?

daniel-montanari (commented):
Looks good overall. Needs the 3.11 features fixed for 3.8.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
+"""
+This module provides the user interface for fixate. It is agnostic of the
+actual implementation of the UI and provides a standard set of functions used
+to obtain or display information from/to the user.
+"""
+
+from typing import Callable, Any
+from queue import Queue, Empty
+from enum import StrEnum
```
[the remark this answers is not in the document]
jcollins1983: Now that we've dropped python prior to 3.12, this won't be an issue.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+        Returns:
+            bool: True if the response is valid, False otherwise
+        """
+        return self.func(resp)
+
+    def __str__(self) -> str:
+        return self.error_msg
+
+
+class UiColour(StrEnum):
```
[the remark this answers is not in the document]
jcollins1983: Now that we've dropped python prior to 3.12, this won't be an issue.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+        Args:
+            resp (Any): The response to validate
+
+        Returns:
+            bool: True if the response is valid, False otherwise
+        """
+        return self.func(resp)
+
+    def __str__(self) -> str:
+        return self.error_msg
```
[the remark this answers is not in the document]
jcollins1983: Will be fixed in an incoming commit.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+    BLUE = "blue"
+    YELLOW = "yellow"
+    WHITE = "white"
+    BLACK = "black"
+    CYAN = "cyan"
+    MAGENTA = "magenta"
+    GREY = "grey"
+
+
+def _user_request_input(msg: str):
```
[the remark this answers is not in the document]
jcollins1983: Fixed in incoming commit. I would have thought that mypy would flag an untyped return.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+
+_ten_digit_int_serial_v = Validator(
+    _ten_digit_int_serial, "Please enter a 10 digit serial number"
+)
+
+
+def user_serial(
+    msg: str,
+    validator: Validator = _ten_digit_int_serial_v,
+    return_type: int | str = int,
```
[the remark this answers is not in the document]
jcollins1983: typing.Type is [deprecated](https://docs.python.org/3.13/library/typing.html#typing.Type). [type](https://docs.python.org/3.13/library/functions.html#type) is the thing to use now.

Mypy wasn't happy with how this was done anyway.

Also, this is the problem with starting something then not touching it for over a year. It's probably 6 of one half a dozen of the other in terms of doing it this way or always returning a string then manually converting later.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+    attempts: int = 5,
+) -> Any:
```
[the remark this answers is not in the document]
jcollins1983: I've fixed this, but using the python 3.10+ syntax

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+        raise ValueError(f"Requires at least two choices to work, {choices} provided")
+    q = Queue()
+    pub.sendMessage("UI_block_start")
+    pub.sendMessage("UI_req_choices", msg=msg, q=q, choices=choices)
+    resp = q.get()
+    pub.sendMessage("UI_block_end")
+    return resp
+
+
+def _choice_from_response(choices: tuple, resp: str) -> str | bool:
```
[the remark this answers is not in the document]
jcollins1983: Fixed in incoming commit.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+    resp = _user_request_input(msg)
+    for _ in range(attempts):
+        if validator(resp):
+            return return_type(resp)
+        pub.sendMessage("UI_display_important", msg=f"Invalid input: {validator}")
+        resp = _user_request_input(msg)
+    raise UserInputError("User failed to enter the correct format serial number")
+
+
+def _user_req_choices(msg: str, choices: tuple):
```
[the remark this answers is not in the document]
jcollins1983: Fixed in incoming commit

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+
+        def set_target_finished_callback(self, callback):
+            self.target_finished_callback = callback
+
+    callback_obj = UserActionCallback()
+    pub.sendMessage("UI_action", msg=msg, callback_obj=callback_obj)
+    try:
+        while True:
+            try:
+                callback_obj.user_cancel_queue.get_nowait()
```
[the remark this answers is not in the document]
jcollins1983: I'm not getting an error when I run mypy...

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+    Args:
+        msg (str): Message to display to the user
+        action_monitor (function): A function that will be called until the user action is cancelled. The function
+            should return False if it hasn't completed. If the action is finished return True.
+
+    Returns:
+        bool: True if the action is finished, False otherwise
+    """
+    # UserActionCallback is used to handle the cancellation of the action either by the user or by the action itself
+    class UserActionCallback:
```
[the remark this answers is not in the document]
jcollins1983: I'm not sure I follow. This is defined in the common module that both cmd and gui use.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+        if resp.startswith(choice[0]):
+            return choice
+    return False
+
+
+def _user_choices(msg: str, choices: tuple, attempts: int = 5) -> str:
+    resp = _user_req_choices(msg, choices).upper()
+    for _ in range(attempts):
+        choice = _choice_from_response(choices, resp)
+        if choice:
```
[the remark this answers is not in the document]
jcollins1983: Correct, I don't think it would make sense to have an empty string as an option.

src/fixate/_ui.py
```diff
@@ -0,0 +1,341 @@
...
+        if validator(resp):
+            return return_type(resp)
+        pub.sendMessage("UI_display_important", msg=f"Invalid input: {validator}")
+        resp = _user_request_input(msg)
+    raise UserInputError("User failed to enter the correct format serial number")
+
+
+def _user_req_choices(msg: str, choices: tuple):
+    # TODO - do we really need this check since this is a private function and any callers should be calling correctly
+    if len(choices) < 2:
```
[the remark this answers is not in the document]
jcollins1983: Agreed, I've switched to be an assert.

src/fixate/core/ui.py
```diff
@@ -55,17 +70,24 @@ def _user_req_choices(msg, choices, target=None, attempts=5):
...
-        q=q,
-        choices=choices,
-        target=target,
-        attempts=attempts,
-    )
-    resp = q.get()
+    for _ in range(attempts):
+        pub.sendMessage("UI_req_choices", msg=msg, q=q, choices=choices)
+        resp = q.get()
+        ret_val = target(resp, choices)
```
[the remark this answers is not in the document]
jcollins1983: As per my previous comment on this topic.

src/fixate/ui_cmdline/cmd_line.py
```diff
@@ -105,14 +107,19 @@ def unregister_cmd_line():
...
     wrapper.initial_indent = first_line_fill
     wrapper.subsequent_indent = subsequent_line_fill
     for ind, line in enumerate(text_str.splitlines()):
         if ind != 0:
             wrapper.initial_indent = subsequent_line_fill
         lines.append(wrapper.fill(line))
+    # reset the indents, calls to this method should not affect the global state
+    wrapper.initial_indent = _wrapper_initial_indent
+    wrapper.subsequent_indent = _wrapper_subsequent_indent
     return "\n".join(lines)
```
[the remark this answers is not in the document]
jcollins1983: What kind of bugs? Not to my knowledge.

src/fixate/core/ui.py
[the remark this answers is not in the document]
jcollins1983: Agree, I don't think it makes much sense for the target to be optional.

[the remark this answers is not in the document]
jcollins1983: Agree, I don't think it makes much sense for the target to be optional.

src/fixate/ui_cmdline/cmd_line.py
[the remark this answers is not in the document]
jcollins1983: It's been a while since I looked at this code, but IIRC if you changed the indent at any point, it would stick around on the next call, hence why the global state was being stashed then recalled after modification.
There is usage of the optional arguments at line 171

src/fixate/ui_gui_qt/ui_gui_qt.py
jcollins1983: Sure, doesn't hurt to be consistent, so I don't recall seeing the same issues in the GUI (but I wouldn't trust my memory from back then)

src/fixate/ui_gui_qt/ui_gui_qt.py
jcollins1983: It's possible that that was added after I started this work. The blame says it's a similar age. I've removed the comments and adjusted the queue additions, we're not using the ("Result", "Message") format anymore.

src/fixate/ui_gui_qt/ui_gui_qt.py
jcollins1983: It took me a second to see what you were talking about... fixed.

jcollins1983 (commented):
I started answering comments in VS Code and it decided that I was doing a review... anyhow... here we are.
I think I've addressed all the things you raised. I've left the conversations as unresolved, as I like to leave that to the reviewer to determine. I also added a file that I used for manual testing of how the UI is displayed.

jcollins1983:
This is ready to be reviewed again.

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+        msg (str): A message that will be shown to the user
+
+    Returns:
+        resp (str): 'YES' or 'NO'
+    """
+    CHOICES = ("YES", "NO")
+    return _user_choices(msg, CHOICES, attempts)
+
+
+def _user_retry_abort_fail(msg: str, attempts: int = 1) -> str:
```
daniel-montanari, line 194: Should this be private? It isn't used here but is imported in other modules. This highlights a different colour in vs code to show it isn't used and makes me want to delete it.

src/fixate/main.py
```diff
@@ -11,7 +11,8 @@
 from pathlib import Path
 import fixate.config
 from fixate.core.exceptions import SequenceAbort
-from fixate.core.ui import user_info_important, user_serial, user_ok
+from fixate.core.ui import user_info_important, user_ok
```
daniel-montanari, line 14: I'm a bit confused at the difference between core.ui and _ui

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+
+T = TypeVar("T")
+
+
+class Validator(Generic[T]):
+    """
+    Defines a validator object that can be used to validate user input.
+    """
+
+    def __init__(self, func: Callable[[T], bool], errror_msg: str = "Invalid input"):
```
daniel-montanari, line 27: typo

src/fixate/main.py
daniel-montanari: Probably shouldn't be using core at all

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+import time
+from pubsub import pub
+
+# going to honour the post sequence info display from `ui.py`
+from fixate.config import RESOURCES
+from fixate.core.exceptions import UserInputError
+from collections import OrderedDict
+
+
+T = TypeVar("T")
```
daniel-montanari, line 19: https://typing.python.org/en/latest/spec/generics.html

src/fixate/ui_cmdline/cmd_line.py
```diff
@@ -242,27 +178,39 @@ def _user_display(msg):
...
     print("!" * wrapper.width)
 
 
+def _user_image_or_gif(path, caller):
+    if caller == "UI_image":
+        disp_str = "image"
+    else:
+        disp_str = "GIF"
+    print("\a")
+    _user_display_important(f"Oh oh, {disp_str} display not supported in command line")
```
daniel-montanari, line 205: Maybe don't need this if we think that showing the link to the image is a valid way of handling this?

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+    for _ in range(attempts):
+        if validator(resp):
+            return return_type(resp)
+        pub.sendMessage("UI_display_important", msg=f"Invalid input: {validator}")
+        resp = _user_request_input(msg)
+    raise UserInputError("User failed to enter the correct format serial number")
+
+
+def _user_req_choices(msg: str, choices: tuple[str, ...]) -> str:
+    assert len(choices) >= 2, "There must be at least 2 choices"
```
daniel-montanari, line 148: Ignore what I said about NATA, it uses user_input to choose a number from a list of options.

daniel-montanari: Image

Looks like the old UI already has this.

There is always potentially the dirty workaround for NATA as an edge case of putting a dummy entry into the system, otherwise if we agree we want this to be a developer warning rather than an error we should change this.

Or push this out to another PR and leave as is here.

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+    pub.sendMessage("UI_block_start")
+    pub.sendMessage("UI_req_choices", msg=msg, q=q, choices=choices)
+    resp = q.get()
+    pub.sendMessage("UI_block_end")
+    return resp
+
+
+def _choice_from_response(choices: tuple[str, ...], resp: str) -> str | Literal[False]:
+    for choice in choices:
+        if resp.startswith(choice[0]):
```
daniel-montanari, line 159: Image

Is this intentionally different to the old code?

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+        resp = _user_request_input(msg)
+    raise UserInputError("User failed to enter the correct format serial number")
+
+
+def _user_req_choices(msg: str, choices: tuple[str, ...]) -> str:
+    assert len(choices) >= 2, "There must be at least 2 choices"
+    q: Queue[str] = Queue()
+    pub.sendMessage("UI_block_start")
+    pub.sendMessage("UI_req_choices", msg=msg, q=q, choices=choices)
+    resp = q.get()
```
daniel-montanari, line 152: Does this have the same bug here https://github.com/PyFixate/Fixate/issues/213

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+def _user_request_input(msg: str) -> str:
+    q: Queue[str] = Queue()
+    pub.sendMessage("UI_block_start")
+    pub.sendMessage("UI_req_input", msg=msg, q=q)
+    resp = q.get()
+    pub.sendMessage("UI_block_end")
+    return resp
+
+
+def user_input(msg: str) -> str:
```
daniel-montanari, line 71: Is it intentional that the retry logic has been removed from this? In the old ui user_input and user_input_float are the same thing with a different validator

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+        msg (str): A message that will be shown to the user
+        attempts (int): Number of attempts the user has to get the input right
+
+    Returns:
+        resp (float): The converted user response from the UI
+
+    Raises:
+        UserInputError: If the user fails to enter a number after the specified number of attempts
+    """
+    resp = _user_request_input(msg)
```
daniel-montanari, line 98: Does it make more sense to move resp into the start of the loop and remove from the except

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+        msg (str): A message that will be shown to the user
+        validator (Validator): An optional function to validate the serial number,
+            defaults to checking for a 10 digit integer. This function shall return
+            True if the serial number is valid, False otherwise.
+        return_type (int | str): The type to return the serial number as, defaults to int
+
+    Returns:
+        resp (str): The user response from the UI
+    """
+    resp = _user_request_input(msg)
```
daniel-montanari, line 138: Same here

src/fixate/_ui.py
```diff
@@ -0,0 +1,344 @@
...
+
+def _choice_from_response(choices: tuple[str, ...], resp: str) -> str | Literal[False]:
+    for choice in choices:
+        if resp.startswith(choice[0]):
+            return choice
+    return False
+
+
+def _user_choices(msg: str, choices: tuple[str, ...], attempts: int = 5) -> str:
+    resp = _user_req_choices(msg, choices).upper()
```
daniel-montanari, line 165: Same here

src/fixate/_ui.py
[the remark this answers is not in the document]
daniel-montanari: I think what I meant was in _topic_UI_action and _user_action to add some typing info on what callback_obj is.

src/fixate/_ui.py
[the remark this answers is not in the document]
daniel-montanari: Looks like this setting isn't turned on for this file. I might've had everything at strict when I wrote this.

I think for this PR I would keep as is, but we probably need to make it more obvious in the project readme how to set these sorts of things up when writing new modules. 
The general intent is over time we fix up errors and add in stricter options.

test/manual/test_ui.py
```diff
@@ -0,0 +1,107 @@
...
+        user_post_sequence_info_fail(
+            "This is post sequence info that should only show if the sequence fails"
+        )
+        chk_fails("You will need to select FAIL or ABORT for this test to work.")
+
+
+PASSING_TESTS = [TestsPassing()]
+FAILING_TESTS = [TestsFailing()]
+
+test_data = {
```
daniel-montanari, line 92: Is there a reason these require an argument and don't play one after the other?

daniel-montanari (commented):
Old tests still pass.
New tests seem ok.

A few comments to talk over but nothing majorly wrong.

src/fixate/_ui.py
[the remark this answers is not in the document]
jcollins1983: Ok, agree. I've added an assert as the developer making tweaks or additions to the UI should be looking after the assignment of `user_cancel_queue` and `target_finished_callback`. Also added typing for these.

src/fixate/_ui.py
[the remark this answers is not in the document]
jcollins1983: In all cases it should be a `UserActionCallback`, but this is defined within the `user_action` method, I don't see an easy way of typing this in the calling functions. I have noticed though that the docstrings for the cmdline was incorrect, so I've updated that.

src/fixate/_ui.py
jcollins1983: It's private in the sense that we don't want to expose it as part of the public interface. Why Pylance doesn't highlight it as used I don't know, it knows where it's used if you ctrl+click on in or show references though.

I have the same tendency to want to delete things that aren't used, since the mechanism used in `__init__.py` is to expose the methods that _are_ part of the public interface, I've removed the `_` so that  Pylance treats it properly, and updated the comment to indicate that it should not be included in the public interface. You still have to explicitly import it from a private module to get to it. I think I'm happy enough with that approach.

src/fixate/main.py
jcollins1983: Yep, there were missed. Now coming from the new implementation.

src/fixate/_ui.py
jcollins1983: Fixed.

src/fixate/_ui.py
jcollins1983: Switched to the alternate method of defining generics. Looks cleaner too IMO.

src/fixate/ui_cmdline/cmd_line.py
jcollins1983: If we're running a script that should display images in the CLI I want it to be clear that the script should probably be run in the GUI.

src/fixate/_ui.py
jcollins1983: Yeah, not sure what I was thinking when I did this. 
With what used to be, you'd get a true response if you responded with Y, YE or YES if the choice was YES, but with what I did you'd get a true response if you responded with the same set but also with any other response starting with Y, like YASS, not really what we want. As I said, not really sure what I was thinking when I did this.

In both cases, if there were multiple choices that started with the same letter, you'd get a hit on the first choice that begins with the entered letter if only one letter was entered, this is only an issue in the CLI, with the GUI you get then entire word for the response. With that in mind, I'll go back to the original logic.

Whilst testing this I found an issue with the retry logic where if an invalid response is entered the user is asked again to input, but if attempts is not overridden to more than 1 this shouldn't happen etc. I've fixed this now too.

src/fixate/_ui.py
jcollins1983: No, `_user_req_choices` only gets a value from the user (in the GUI shows buttons to press and gets the value for the button that was pressed). It's up to the caller to process that response. Currently the only caller is `_user_choices` which will either get a valid choice from the input or return False (which is what would happen if the user entered 0 or any other falsey value), in which case either another attempt is allowed or an exception is raised.

Currently there is no "gerneric" form of user_choices (the public version of this was removed in 0.5.2 - see line 204 of the release-notes), so there is currently no mechanism for a script to introduce a potentially problematic set of choices, though I have tested with a use of the private `_user_choices` and it happily accepts any string including an empty string and will return that as the choice if that is entered. It currently doesn't accept any numeric values.

Also, I made a point of addressing #213 in this PR.

src/fixate/_ui.py
jcollins1983: It hasn't, `user_input` from the original `ui` module has no retry logic. It's just used to get a raw input from the user. 

In general, I've made anything that gets user input just that, a mechanism for getting input. If a more specific function uses the function to get input and there is any processing to be done, this is done by the caller, including any retry logic.

src/fixate/_ui.py
jcollins1983: Yes, this will fix the bug that I mentioned earlier in `_user_choices` if attempts = 1.

src/fixate/_ui.py
jcollins1983: Ditto.

src/fixate/_ui.py
jcollins1983: Yep, found it before I go to this comment. I would much prefer GH comments to be linear with respect to the code instead of temporally linear (they probably would be in a different view... 🤷‍♂️)

test/manual/test_ui.py
jcollins1983: Yes, the post sequence info is dependent on the overall result of the test.

src/fixate/_ui.py
jcollins1983: Except now it looks like I will need to bump the version of black to get it through

Image

jcollins1983:
@daniel-montanari I think I've addressed all of your comments. There were some outdated comments that I marked as resolved, but there are still a few comments where I responded and in general, I prefer the reviewer to mark as resolved if they're happy with the response - so I've left them for you to mark as resolved if you're happy.

src/fixate/_ui.py
jcollins1983: And now that I've done that _and_ with tox doing its thing (which I didn't run locally before pushing because dumb that's why) the CI run is failing because of other things that the newer version of black wants to introduce. This will be a later problem now.

jcollins1983: Ok, so this rang a bell from when I was doing the ftdi driver update that is being put into a separate repo under PyFixate (coming soonTM). @clint-lawrence and I spoke about either getting rid of the tox black environment or running pre-commit in tox as well.
I think after speaking to @daniel-montanari about this also, I'm leaning towards just removing the black section from tox. This means that we won't have bulk formatting changes coming in on an ostensibly unrelated PR. If any unchanged files are touched in the future, then these will be caught up with the newer black enforcements.

src/fixate/_ui.py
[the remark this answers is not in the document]
daniel-montanari: Looks good.

src/fixate/ui_cmdline/cmd_line.py
```diff
@@ -125,15 +131,13 @@ def _user_action(msg, callback_obj):
     system is automatically monitoring for success but has no way of monitoring failure.
     :param msg:
      Information for the user
-    :param q:
-     The queue object to put false if the user fails the test
-    :param abort:
-     The queue object to abort this monitoring as the test has already passed.
+    :param callback_obj:
+     The callback object to handle user canncellation or target completion. This is a UserActionCallback object as defined in fixate._ui.user_action.
```
daniel-montanari: canncellation

src/fixate/_ui.py
daniel-montanari: Isn't there retry logic buried in `_user_req_input`? Or does it not trigger for `user_input` in the `ui` module?

I agree though with moving retry logic up a level rather than having it buried like this.

src/fixate/_ui.py
daniel-montanari: Sounds good.

src/fixate/_ui.py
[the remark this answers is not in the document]
daniel-montanari: Sounds good

src/fixate/ui_gui_qt/ui_gui_qt.py
```diff
@@ -716,7 +721,7 @@ def _topic_UI_action(self, msg, callback_obj):
         monitoring failure.
         :param msg: Information for the user
         :param callback_obj:
-         callback_obj used to communicate back to the user_action call in ui.py.
+         The callback object to handle user canncellation or target completion. This is a UserActionCallback object as defined in fixate._ui.user_action.
```
daniel-montanari: canncellation

src/fixate/_ui.py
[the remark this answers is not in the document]
daniel-montanari: I think thats what I was getting at.
Other modules use the callback so it probably shouldn't live inside `user_action`. Since they aren't checked by mypy there isn't too much benefit to the type hinting I was originally talking about.

src/fixate/ui_gui_qt/ui_gui_qt.py
daniel-montanari: look good

src/fixate/ui_cmdline/cmd_line.py
daniel-montanari: ok

src/fixate/ui_cmdline/cmd_line.py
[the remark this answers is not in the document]
daniel-montanari: Ah yeah I see now, so you've moved where the indents got reset out of `_user_input` up to here.
We could probably keep simplifying this module to remove the global state but for now the bug fix is fine.

test/manual/test_ui.py
daniel-montanari: ok

.github/workflows/test.yml
```diff
@@ -32,9 +32,6 @@ jobs:
     - name: Run tests with tox
       run: |
         tox -e py
-    - name: Check formatting
```
daniel-montanari: Is pre-commit being used now instead of tox for running black?
I can't see any job for formatting in the last CI run.

daniel-montanari (changes requested):
Check CI, minor comments otherwise.
Will need to pull latest changes into this to fix failing tests.

.github/workflows/test.yml
jcollins1983: I refer you to this comment @daniel-montanari 
https://github.com/PyFixate/Fixate/pull/214#discussion_r3293861338

jcollins1983:
> Check CI, minor comments otherwise. Will need to pull latest changes into this to fix failing tests.

What failing tests? And what specifically in the CI? Vague comments don't help in the resolution of these things...

src/fixate/_ui.py
jcollins1983: _user_req_input is in the old UI _only_. Keep in mind that the old UI will be phased out. The stuff that really matters is in _ui.py.

src/fixate/_ui.py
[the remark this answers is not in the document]
jcollins1983: The callback is called from other modules, that's what callbacks are designed for. The callback is specifically for handling cancellation of `user_action` nothing else, so I see no reason for it _not_ to be where it is.

src/fixate/ui_cmdline/cmd_line.py
jcollins1983: Fixed

src/fixate/ui_gui_qt/ui_gui_qt.py
jcollins1983: Fixed

jcollins1983:
@daniel-montanari I've addressed the conflict, tests are passing. There are a whole bunch of unresolved comments in outdated sections that I would like you to either provide further comment on or resolve if you're happy.

With regard to the formatting. It is currently reliant upon pre-commit and the person doing any work having the pre-commit hooks installed. I'm open to including pre-commit in the actions to use the same configuration that each person should be using when making changes in a branch. The reason I left it out is that when we need to bump versions of black etc. to accommodate for new features required for the code we're modifying we don't end up with the entire codebase being updated "just because", I would prefer to defer this to when/if this code needs to be modified in the future, but if there's enough desire from others to continue to format the entire codebase each time the formatting tools are updated then I'll acquiesce.

src/fixate/_ui.py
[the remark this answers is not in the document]
daniel-montanari: Ok

src/fixate/_ui.py
daniel-montanari: ok

src/fixate/main.py
daniel-montanari: ok

src/fixate/_ui.py
daniel-montanari: ok

src/fixate/_ui.py
daniel-montanari: Bumped black and reran https://github.com/PyFixate/Fixate/pull/261
I found this pretty easy to merge into my vmux branch. The only conflicts I had were the config files I had touched and my code that had not been run through the new version of black.

If this ends up being a pain to merge for you then I'm happy to abandon that PR. Really it just means a few extra whitespace diffs when someone touches a file that hasn't been formatted in a while.

src/fixate/_ui.py
daniel-montanari: Yep

src/fixate/_ui.py
daniel-montanari: ok

src/fixate/_ui.py
daniel-montanari: Yeah, it makes it a pain when running through the code, but has saved me when I split thoughts across messages.
I do need to make comments make sense without needing the context of other comments.

src/fixate/_ui.py
daniel-montanari: ok

.github/workflows/test.yml
daniel-montanari: Replied to that comment

src/fixate/_ui.py
daniel-montanari: I think with either case using startswith will result in some edge cases rather than an exact match.
Until then sticking with the old logic but fixing the bug should be good.

daniel-montanari:
> > Check CI, minor comments otherwise. Will need to pull latest changes into this to fix failing tests.
> 
> What failing tests? And what specifically in the CI? Vague comments don't help in the resolution of these things...

I should've been more specific.
There were some tests failing due to breaking changes in packages. They would've only happened if you reran the CI.

daniel-montanari (commented):
Will approve pending your decision of what to do with the formatting.
Can either have a go merging my changes in or abandon and go with these changes.
https://github.com/PyFixate/Fixate/pull/214#discussion_r3293861338
https://github.com/PyFixate/Fixate/pull/261

daniel-montanari (approved):
Looks good

PR #215, opened 2024-09-01
fix workflow branch name master -> main

jcollins1983:


clint-lawrence (approved):
Yeah, that's my bad. Looks good to me!

PR #216, opened 2024-11-15
DMM Drivers - NPLC and Min/Max/Avg support

Overlord360:
Add ability to configure NPLC on a driver level in both the Fluke and Keithley DMM drivers

Added a function to have the drivers sample x times and then return the minimum, maximum and average of the samples

### DIFF
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -116,7 +116,7 @@ class Fluke8846A(DMM):
         automatically samples the DMM for a given number of samples and returns the min, max, and average values
         :param samples: number of samples to take
         :param sample_time: time to wait for the DMM to take the samples
-        return: min, avg, max values as floats
+        return: min, avg, max values as floats in a dictionary
         """
 
         self._write(f"SAMP:COUN {samples}")
@@ -128,7 +128,9 @@ class Fluke8846A(DMM):
         avg_ = self.instrument.query_ascii_values("CALC:AVER:AVER?")[0]
         max_ = self.instrument.query_ascii_values("CALC:AVER:MAX?")[0]
 
-        return min_, avg_, max_
+        values = {"min": min_, "avg": avg_, "max": max_}
+
+        return values
 
     def reset(self):
         """
@@ -359,24 +361,3 @@ class Fluke8846A(DMM):
         # Remove the CONF: from the start of the string
         mode_str = mode_str.replace("CONF:", "")
         return float(self.instrument.query(f"{mode_str}:NPLC?"))
-
-    # context manager for setting NPLC
-    class _nplc_context_manager(object):
-        def __init__(self, dmm, nplc=None):
-            self.dmm = dmm
-            self.nplc = nplc
-            self.original_nplc = self.dmm.get_nplc()
-
-        def __enter__(self):
-            self.dmm.set_nplc(self.nplc)
-
-        # return to default NPLC setting
-        def __exit__(self, exc_type, exc_val, exc_tb):
-            # check if an exception was raised
-            if exc_type is not None or exc_val is not None or exc_tb is not None:
-                return False  # re-raise the exception
-            # continue with the exit process
-            self.dmm.set_nplc(self.original_nplc)
-
-    def nplc(self, nplc=None):
-        return self._nplc_context_manager(self, nplc)
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -111,6 +111,25 @@ class Fluke8846A(DMM):
         with self.lock:
             return self._read_measurements()
 
+    def min_avg_max(self, samples=1, sample_time=1):
+        """
+        automatically samples the DMM for a given number of samples and returns the min, max, and average values
+        :param samples: number of samples to take
+        :param sample_time: time to wait for the DMM to take the samples
+        return: min, avg, max values as floats
+        """
+
+        self._write(f"SAMP:COUN {samples}")
+        self._write("CALC:FUNC AVER")
+        self._write("CALC:STAT ON")
+        self._write("INIT")
+        time.sleep(sample_time)
+        _min = self.instrument.query_ascii_values("CALC:AVER:MIN?")[0]
+        _avg = self.instrument.query_ascii_values("CALC:AVER:AVER?")[0]
+        _max = self.instrument.query_ascii_values("CALC:AVER:MAX?")[0]
+
+        return _min, _avg, _max
+
     def reset(self):
         """
         Checks for errors and then returns DMM to power up state
@@ -233,14 +252,13 @@ class Fluke8846A(DMM):
         self._is_error()
 
     def set_nplc(self, nplc=None, reset=False):
-        if nplc not in self._nplc_settings:
+        if reset is True or nplc is None:
+            nplc = self._default_nplc
+        elif nplc not in self._nplc_settings:
             raise ParameterError(f"Invalid NPLC setting {nplc}")
 
         if self._mode not in self._nplc_modes:
             raise ParameterError(f"NPLC setting not available for mode {self._mode}")
-
-        if reset is True or nplc is None:
-            nplc = self._default_nplc
 
         mode_str = f"{self._modes[self._mode]}"
 
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -128,7 +128,7 @@ class Fluke8846A(DMM):
         avg_ = self.instrument.query_ascii_values("CALC:AVER:AVER?")[0]
         max_ = self.instrument.query_ascii_values("CALC:AVER:MAX?")[0]
 
-        values = {"min": min_, "avg": avg_, "max": max_}
+        values = DMM.MeasurementStats(min=min_, avg=avg_, max=max_)
 
         return values
 
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -50,7 +50,7 @@ class Fluke8846A(DMM):
             "temperature",
             "ftemperature",
         ]
-        self._nplc_settings = [0.02, 0.2, 1, 10, 100]
+        self._nplc_settings = [0.02, 0.2, 1, 10]
         self._default_nplc = 10  # Default NPLC setting as per Fluke 8846A manual
         self._init_string = ""  # Unchanging
 
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -130,6 +130,10 @@ class Fluke8846A(DMM):
 
         values = DMM.MeasurementStats(min=min_, avg=avg_, max=max_)
 
+        # clean up
+        self._write("CALC:STAT OFF")
+        self._write("SAMP:COUN 1")
+
         return values
 
     def reset(self):
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -243,9 +243,9 @@ class Fluke8846A(DMM):
             nplc = self._default_nplc
 
         mode_str = f"{self._modes[self._mode]}"
-        mode_str = mode_str.replace(
-            "CONF:", ""
-        )  # Remove the CONF: from the start of the string
+
+        # Remove the CONF: from the start of the string
+        mode_str = mode_str.replace("CONF:", "")
 
         self._write(f"{mode_str}:NPLC {nplc}")  # e.g. VOLT:DC:NPLC 10
 
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -116,7 +116,7 @@ class Fluke8846A(DMM):
         automatically samples the DMM for a given number of samples and returns the min, max, and average values
         :param samples: number of samples to take
         :param sample_time: time to wait for the DMM to take the samples
-        return: min, avg, max values as floats in a dictionary
+        return: min, avg, max values as floats in a dataclass
         """
 
         self._write(f"SAMP:COUN {samples}")
--- a/src/fixate/drivers/dmm/fluke_8846a.py
+++ b/src/fixate/drivers/dmm/fluke_8846a.py
@@ -42,6 +42,16 @@ class Fluke8846A(DMM):
             "continuity": "CONF:CONTinuity",
             "diode": "CONF:DIODe",
         }
+        self._nplc_modes = [
+            "resistance",
+            "fresistance",
+            "voltage_dc",
+            "current_dc",
+            "temperature",
+            "ftemperature",
+        ]
+        self._nplc_settings = [0.02, 0.2, 1, 10, 100]
+        self._default_nplc = 10  # Default NPLC setting as per Fluke 8846A manual
         self._init_string = ""  # Unchanging
 
     @property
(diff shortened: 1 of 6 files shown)


src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -101,6 +111,25 @@ def measurements(self):
         with self.lock:
             return self._read_measurements()
 
+    def min_avg_max(self, samples=1, sample_time=1):
```
Jasper-Harvey0, line 114: I personally don't think we should be adding math functions into the DMM driver. We already have the ability to take a whole bunch of measurements, we can do whatever statistical analysis we want in the test script. 
I am happy to be overruled on this by @clint-lawrence or @jcollins1983 but that's just my opinion. 

The way we currently have to sample a bunch of measurements uses the manual triggering functions:
```
dmm.set_manual_trigger(samples=100) # Get ready to take 100 samples
dmm.trigger() # Start taking samples
values = dmm.measurements() # Get a list of the measurements
```
I have realised that the ``trigger()`` function (at least for the Keithley) will give a timeout error if you set a very large number of samples and don't increase ``instrument.timeout``. Maybe we just need to improve that function to wait for the samples to be taken, by querying some status register on the DMM.

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -40,7 +40,16 @@ def __init__(self, instrument, *args, **kwargs):
...
         }
-
+        self._nplc_modes = [
+            "voltage_dc",
+            "current_dc",
+            "resistance",
+            "fresistance",
+            "diode",
+            "temperature",
+        ]
```
Jasper-Harvey0, line 50: The Keithley driver does not have ``self._nplc_settings``. Does this mean I can pick any float between 0.0005 and 12?

I can see this becoming an issue when someone builds a test script with the Keithley and sets nplc = 3 for example. This will cause the Fluke to error because 3 is not an available setting. 
This is very much related to the range settings problem I had a think about earlier in the year. https://github.com/PyFixate/Fixate/pull/205

Jasper-Harvey0 (changes requested):

src/fixate/drivers/dmm/keithley_6500.py
Overlord360: That is correct. This is why I have error checking in the fluke DMM driver set_NPLC function to check that it is a specific setting where as in the keithley it just checks that it is within the instruments range. 

This will mean that if you try to set NPLC to 3, the keithley driver will run fine, but the fluke driver will throw a parameter error (and not try and set NPLC to an invalid setting)

src/fixate/drivers/dmm/fluke_8846a.py
clint-lawrence: > I personally don't think we should be adding math functions into the DMM driver

I'm generally in agreement. Unless it someone we do lots of and perhaps its worth the effort to standardise. But in that case, have a generic function shared by the different driver would be preferable to having to implement on different drivers.

Other option open to us are:
- Make sure any common patterns are added to documentation, so they don't need to be re-discovered
- Create a module of utility functions that capture common pattern, but aren't tied to a specific driver

> I have realised that the trigger() function (at least for the Keithley) will give a timeout error if you set a very large number of samples and don't increase instrument.timeout

This has been a pain in the past. As best I can tell, the semantics around how to actually use SCPI command efficiently is inconsistent across instruments and poorly documented. I'm sure there is at least one place where we adjust the timeout on the fly to achieve this.

Overlord360: I tried using the trigger() function in the keitley driver for this and did run into that timeout issue.

The way I think about it is that the DMMs have the ability to do these calculations in the device so it makes sense to be able to use them. It's not that we're doing the calculations in python in the driver.

src/fixate/drivers/dmm/keithley_6500.py
clint-lawrence: Unless there is a compelling reason driven by test requirements, is there a reasonable lowest common denominator we support here, even if that means the full range of values isn't available

jcollins1983: Ideally, a script wouldn't need to be tweaked if we swap between Fluke and Keithly. So, perhaps we limit to the intersection of the capabilities of each instrument?
.e.g. `[0.02, 0.2, 1, 10]`

src/fixate/drivers/dmm/fluke_8846a.py
jcollins1983: >> I personally don't think we should be adding math functions into the DMM driver

>I'm generally in agreement. Unless it someone we do lots of and perhaps its worth the effort to standardise. But in that case, have a generic function shared by the different driver would be preferable to having to implement on different drivers.

I agree with both statements here.

Overlord360: > > > I personally don't think we should be adding math functions into the DMM driver
> 
> > I'm generally in agreement. Unless it someone we do lots of and perhaps its worth the effort to standardise. But in that case, have a generic function shared by the different driver would be preferable to having to implement on different drivers.
> 
> I agree with both statements here.

The motivator behind making this change is to simplify the test script drastically as well as to allow it to work on both the keithley and fluke.
image

clint-lawrence: @Overlord360 Thanks, that is a great example to add to the PR description, so reviewers understand the context a little better. I was trying to avoid getting too sucked into reviewing this, but since I'm here a few thoughts...

`set_nplc` would make a good context manager
```python
with dmm.set_nplc(nplc):
    return dmm.measure()
```
If you don't like that, then perhaps return the old value from `set_nplc`. Then it can be stored locally and passed back into the second call. That is a fairly common pattern for API that changes/reset some kind of global state. But a context manager is better because it is more reliable in the event of exception being raise

Avoid prefixing builtin names with an underscore. That convention is typically used to indicate something as private in python. Instead you can use a postfix underscore.

Or even better, return a class with appropriate attributes. Saves the awkward naming and is easier to consume

```python
# in some common file
@dataclass
class MeasurementStats:
    min: float
    max: float
    mean: float
```

Then you can return that class directly from the dmm function.

See the example and understanding that this tidies up existing usage, I'm happy to see this get included.

src/fixate/drivers/dmm/keithley_6500.py
Overlord360: > Ideally, a script wouldn't need to be tweaked if we swap between Fluke and Keithly. So, perhaps we limit to the intersection of the capabilities of each instrument? .e.g. `[0.02, 0.2, 1, 10]`

This has been done as of 4d00493

src/fixate/drivers/dmm/fluke_8846a.py
Overlord360: I've added a context manager for setting nplc in helper.py. 

I've also made the min_avg_max function return a dictionary of the results with the keys: "min", "avg" and "max" as the class that Clint suggested is pretty much just a dictionary.

I've also added some tests for these new functions.

clint-lawrence: > I've also made the min_avg_max function return a dictionary of the results with the keys: "min", "avg" and "max" as the class that Clint suggested is pretty much just a dictionary.

No, I disagree with this. There are a couple of important differences.

It's useful to think of a dataclass as a "record" or "struct" data type; A grouping of members. The significant thing is we know what the members are (`min`, `max` and `avg`). The makes it more discoverable and easier to use at the call site as most IDEs will be able to analyse the code and know the return type of the method and the available attributes on that type. With a dictionary, the keys are completely unconstrained.

The second difference (which is mostly a consequence of the first) is that returning dict with a of known keys is uncommon in python code. While that shouldn't rule it out from consideration, to go against the common practise I expect there to be a strong justification.

One minor thing that I did above but didn't specifically call out. I replaced `avg` with `mean` since it is more specific. I'm not too bother if it stays as `arg`, since it is fairly common for people to assume mean when the method of calculating an average isn't specified.

src/fixate/drivers/dmm/helper.py
```diff
@@ -101,3 +104,30 @@ def reset(self):
...
+            self.nplc = nplc
+            self.original_nplc = dmm.get_nplc()
+
+        def __enter__(self):
+            self.dmm.set_nplc(self.nplc)
+
+        # return to default NPLC setting
+        def __exit__(self, exc_type, exc_val, exc_tb):
+            # check if an exception was raised
+            if exc_type is not None or exc_val is not None or exc_tb is not None:
```
Jasper-Harvey0, line 127: Pretty sure this guy can be reduced to:
```
if exc_type:
```

Overlord360: Done as of 1fdc57f

clint-lawrence: Shouldn't it be removed completely?

I think we want to restore the NPLC setting if there was an error and in normal operation

Jasper-Harvey0 (changes requested):
Not done yet. I will come back in a little bit

src/fixate/drivers/dmm/helper.py
Overlord360: We have to return false otherwise the context manager "swallows" the exception and fixate will not stop executing the test script.

however we could move the `self.dmm.set_nplc(self.original_nplc)` above this check. However, what happens if the reason an exception was thrown was because of an instrument error with the DMM?

test/drivers/test_fluke_8846A.py
```diff
@@ -316,6 +316,78 @@ def test_measurement_diode(funcgen, dmm, rm):
...
+def test_min_avg_max(dmm, rm, funcgen):
+    dmm.voltage_dc()
+    dmm.set_nplc(nplc=0.02)
+    values = dmm.min_avg_max(995, 1.1)
+    min_val = values.min
+    avg_val = values.avg
+    max_val = values.max
+
+    # does not guarantee that there is any input to the DMM so we can't guarantee that the min, avg, max are different
+    assert min_val <= avg_val <= max_val
```
Jasper-Harvey0, line 371: This test is checking the ``min_avg_max()`` function twice. Once with floating inputs to the DMM, and once with the function generator connected. 
We really should just test this once with the function generator. Maybe change the amplitude of the function generator. 

The other thing is, we don't test this function with different modes, or nplc values. You can use the pytest @pytest.mark.parametrize to inject a range of difference test data to the same function.

test/drivers/test_fluke_8846A.py
```diff
@@ -316,6 +316,78 @@ def test_measurement_diode(funcgen, dmm, rm):
...
 
+@pytest.mark.drivertest
+def test_get_nplc(dmm):
+    query = dmm.get_nplc()
+    assert query == pytest.approx(10)
+
+
+@pytest.mark.drivertest
+def test_set_nplc(dmm):
+    dmm.voltage_dc()
```
Jasper-Harvey0, line 327: You can use "@pytest.mark.parametrize" to inject a range of different test data into the function. That will help you test a range of different modes.

src/fixate/drivers/dmm/helper.py
clint-lawrence: > We have to return false otherwise the context manager "swallows" the exception and fixate will not stop executing the test script.

I don't think that is quite correct. The docs say:

> If an exception is supplied, and the method wishes to suppress the exception (i.e., prevent it from being propagated), it should return a true value.

If we want to exception to be re-raised, we simply can do nothing and the implicit return value of `None` will result in the exception being propagated.

> However, what happens if the reason an exception was thrown was because of an instrument error with the DMM?

In this case we will just get a chained exception. https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

src/fixate/drivers/dmm/helper.py
```diff
@@ -101,3 +104,30 @@ def reset(self):
...
+        min: float
+        max: float
+        avg: float
+
+    # context manager for setting NPLC
+    class _nplc_context_manager(object):
+        def __init__(self, dmm, nplc=None):
+            self.dmm = dmm
+            self.nplc = nplc
+            self.original_nplc = dmm.get_nplc()
```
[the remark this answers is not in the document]
Overlord360: This has been done as of b781b40

src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -101,6 +111,27 @@ def measurements(self):
...
+        """
+        automatically samples the DMM for a given number of samples and returns the min, max, and average values
+        :param samples: number of samples to take
+        :param sample_time: time to wait for the DMM to take the samples
+        return: min, avg, max values as floats in a dictionary
+        """
+
+        self._write(f"SAMP:COUN {samples}")
```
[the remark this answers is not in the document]
Overlord360: After diving into the fluke manual i can confirm it does not get reset until you change measurement modes.
I've added the cleanup calls as of 4da962e

test/drivers/test_fluke_8846A.py
Overlord360: I've parameterized the tests as of f657fa2

test/drivers/test_fluke_8846A.py
Overlord360: I've parameterized the tests as of f657fa2

src/fixate/drivers/dmm/helper.py
Overlord360: I've removed the error checking as of 43b8675

src/fixate/drivers/dmm/fluke_8846a.py
Overlord360: This has been done as of 4bd69b0

Jasper-Harvey0 (changes requested):
I think I am happy with this after you do the changes.
Please update the release notes. You will have to pull in the recent changes from main I think?

@clint-lawrence @jcollins1983 Are you two happy to proceed?

test/drivers/test_fluke_8846A.py
```diff
@@ -316,6 +316,192 @@ def test_measurement_diode(funcgen, dmm, rm):
     assert meas == pytest.approx(TEST_DIODE, abs=TEST_DIODE_TOL)
 
 
+@pytest.mark.parametrize(
+    "mode",
+    [
+        ("voltage_dc"),
+        ("current_dc"),
+        ("resistance"),
```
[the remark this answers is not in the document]
Overlord360: Looking at the datasheet for the fluke, it does not support changing nplc on diode measurements (while the keithley does). I could add it to the fluke tests with the xfail mark if we want that coverage.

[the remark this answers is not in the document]
clint-lawrence: What about removing it from the keithley, from the point of view of keeping the two driver as compatible as possible (until we having compelling reason to do otherwise)

clint-lawrence:
> @clint-lawrence @jcollins1983 Are you two happy to proceed?

Yep, my earlier comment have all been address.

src/fixate/drivers/dmm/fluke_8846a.py
```diff
@@ -101,6 +111,31 @@ def measurements(self):
         with self.lock:
             return self._read_measurements()
 
+    def min_avg_max(self, samples=1, sample_time=1):
+        """
+        automatically samples the DMM for a given number of samples and returns the min, max, and average values
+        :param samples: number of samples to take
+        :param sample_time: time to wait for the DMM to take the samples
+        return: min, avg, max values as floats in a dictionary
```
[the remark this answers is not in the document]
Overlord360: done as of 52a49ec

src/fixate/drivers/dmm/keithley_6500.py
```diff
@@ -138,6 +147,33 @@ def measurements(self):
         with self.lock:
             return self._read_measurements()
 
+    def min_avg_max(self, samples=1, sample_time=1):
+        """
+        automatically samples the DMM for a given number of samples and returns the min, max, and average values
+        :param samples: number of samples to take
+        :param sample_time: time to wait for the DMM to take the samples
+        return: min, avg, max values as floats in a dictionary
```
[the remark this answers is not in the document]
Overlord360: done as of 52a49ec

test/drivers/test_fluke_8846A.py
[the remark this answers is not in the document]
Overlord360: I've removed diode as a valid mode for which NPLC can be changed from the keithley driver.
I've also updated the tests accordingly
this is done as of 5d96f54

Jasper-Harvey0 (approved):
Looks good to me

Jasper-Harvey0:
Update Release Notes please!

Jasper-Harvey0 (changes requested):
Update release notes please!

Jasper-Harvey0 (approved):
Looks good.

PR #219, opened 2025-01-13
Release 0.6.4

Jasper-Harvey0:
I would like to get a fixate release done so I can incorporate the changes into Amptest by the end of the month.

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -5,7 +5,7 @@ Release Notes
 *************
 Version 0.6.4
 *************
-Release Date xx/xx/24
+Release Date 31/01/25
 
 New Features
 ############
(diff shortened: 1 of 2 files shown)


docs/release-notes.rst
```diff
@@ -5,7 +5,7 @@ Release Notes
 *************
 Version 0.6.4
 *************
-Release Date xx/xx/24
+Release Date 31/01/25
```
clint-lawrence, line 8: Did you mean "13" and not "31"

Jasper-Harvey0: In my head I was thinking we would have a bunch of dependency updates that I would be sifting through, so I set the date to EOM, and the PR to draft. 
Doesn't look like there are any updates, so I think we are good to go.

PR #220, opened 2025-01-14
Fix issue where a non-boolean variable passed into a testclass would skip the test

Overlord360:
In a test script we were accidentally passing a variable that was not consumed by the function and therefore was passed to the parent class (the testclass) this would then be interpreted as the "skip" variable. The check in sequencer.py does not check if the value is a boolean.

This could result in a code mistake in a test script "silently" skipping a test without being noticed. Unless you go look at the logs.

image
image


jcollins1983 (approved):
Looks like a sensible change.

PR #224, opened 2025-06-26
Update the abort sequence logic to catch exceptions in exit.

Jasper-Harvey0:
The test exit functionality currently does not catch exceptions raised in the exit() function when aborting a sequence early.
This causes the GUI to hang and become unresponsive, meaning you have to force close the window. 

The update now catches exceptions raised in the test exit functions and logs them. I have also removed the try / finally block around run_once() as there is already a lot of try except logic in that function, and it just returns when an exception is caught anyway.

I don't know if we want to catch certain errors, instead of "Exception", but I figure something bad has already happened if we are running that code, more errors should just be logged for processing later?

### DIFF
--- a/src/fixate/sequencer.py
+++ b/src/fixate/sequencer.py
@@ -7,6 +7,9 @@ from fixate.core.exceptions import SequenceAbort, CheckFail
 from fixate.core.ui import user_retry_abort_fail
 from fixate.core.checks import CheckResult
 from fixate.reporting import CSVWriter
+import logging
+
+logger = logging.getLogger(__name__)
 
 STATUS_STATES = ["Idle", "Running", "Paused", "Finished", "Restart", "Aborted"]
 
@@ -219,14 +222,19 @@ class Sequencer:
         """
         self.reporting_service.install()
         self.status = "Running"
-        try:
-            self.run_once()
-        finally:
-            while self.context:
-                top = self.context.top()
-                if isinstance(top.current(), TestList):
+
+        self.run_once()
+
+        while self.context:
+            # Test sequence aborted early for some reason
+            # Run test exit functions
+            top = self.context.top()
+            if isinstance(top.current(), TestList):
+                try:
                     top.current().exit()
-                self.context.pop()
+                except Exception as e:
+                    logger.exception(e)
+            self.context.pop()
 
         self.reporting_service.uninstall()
 


jcollins1983 (commented):
I think it looks reasonable.
Release notes?

Jasper-Harvey0:
Right so I have some basic tests implemented. They just check for an expected set of calls to the pubsub topics, as well as a finish result from the sequencer.
Happy for comments about adding more tests, or making them better.

Jasper-Harvey0:
Ok, I think I have the tests in a slightly better place. They check for correct pubsub calls, as well as correct calls to the test list / test class functions. I've tried to capture the case of having an error in each of the key test phases.

test/core/test_sequencer.py
```diff
@@ -1,19 +1,35 @@
...
+
+    def test(self):
+        self.mock.test_test(self.num)
 
 
-class Lst(FixateTL):
+class MockTestList(TestList):
     """
-    Mock Test List
+    Test class that allows tracing of function calls
```
jcollins1983, line 32: Technically a class, but can you please disambiguate from a Test Class since this is a TestList? Is this just a copy paste from before and you meant to say "Test list that allows ..."?

test/core/test_sequencer.py
```diff
@@ -1,19 +1,35 @@
-import time
-import unittest
-from unittest.mock import MagicMock, call
-from pubsub import pub
+import pytest
 import fixate
-from fixate.core.common import TestList as FixateTL, TestClass as FixateTC
+from fixate.core.common import TestList, TestClass
+from fixate.core.checks import chk_fails, chk_passes
+from fixate.core.exceptions import SequenceAbort
```
jcollins1983, line 5: You've imported `SequenceAbort` but you haven't used it.

jcollins1983 (commented):
I think the tests look good. Just a couple of minor things.

test/core/test_sequencer.py
Jasper-Harvey0: Removed.

test/core/test_sequencer.py
Jasper-Harvey0: Comment updated now.

docs/release-notes.rst
```diff
@@ -2,6 +2,20 @@
...
+*************
+Release Date xx/xx/xx
+
+New Features
+############
+
+Improvements
+############
+
+- Sequencer logic now handles exceptions raised on sequence abort. GUI will no longer hang when a test raises and exception during a test abort.
```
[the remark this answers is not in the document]
Jasper-Harvey0: Yep. Updated.

clint-lawrence:
Anything holding this back? Looks like @jcollins1983 was pretty happy with it, so do we just need the release notes fixed and then merge?

Jasper-Harvey0:
Release notes updated and main has been merged in. I think this is good to go.

Jasper-Harvey0:
@jcollins1983 or @clint-lawrence Good to go?

PR #228, opened 2025-09-05
Fix crashing the SPD3303X processor when saving

Riley-King:
The command `*SAV <X>` takes ~50 ms on the SPD3303X but we only wait for ~21 ms before trying to use it which causes it to crash and needs to be power cycled before it will work again.

Changes;
 - Added a slow write function.
 - `save.*` now uses the slow write function.

### DIFF
--- a/src/fixate/drivers/pps/siglent_spd_3303X.py
+++ b/src/fixate/drivers/pps/siglent_spd_3303X.py
@@ -24,11 +24,11 @@ class SPD3303X(PPS):
 
         self.api = [
             # Save commands
-            ("save.group1", self.write, "*SAV 1"),
-            ("save.group2", self.write, "*SAV 2"),
-            ("save.group3", self.write, "*SAV 3"),
-            ("save.group4", self.write, "*SAV 4"),
-            ("save.group5", self.write, "*SAV 5"),
+            ("save.group1", self._write_slow, "*SAV 1"),
+            ("save.group2", self._write_slow, "*SAV 2"),
+            ("save.group3", self._write_slow, "*SAV 3"),
+            ("save.group4", self._write_slow, "*SAV 4"),
+            ("save.group5", self._write_slow, "*SAV 5"),
             # Recall commands
             ("recall.group1", self.write, "*RCL 1"),
             ("recall.group2", self.write, "*RCL 2"),
@@ -203,6 +203,18 @@ class SPD3303X(PPS):
         for cmd in data.split(";"):
             self.instrument.write(cmd)
             time.sleep(0.02 + len(cmd) / 6000)
+        self._is_error()
+
+    def _write_slow(self, base_str, *args, **kwargs):
+        """
+        The SPD3303X is very slow at saving the current configuration to memory
+        To correct this, an extra 30 ms is added to the delay between commands.
+        This function should be identical to _write with the exception of an additional 0.03 s in wait time
+        """
+        cmds = self._format_string(base_str, **kwargs)
+        for cmd in cmds.split(";"):
+            self.instrument.write(cmd)
+            time.sleep(0.02 + len(cmd) / 6000 + 0.03)
         self._is_error()
 
     @staticmethod


src/fixate/drivers/pps/siglent_spd_3303X.py
```diff
@@ -205,6 +205,18 @@ def _write(self, data):
...
+        The SPD3303X is very slow at saving the current configuration to memory
+        To correct this, an extra 30 ms is added to the delay between commands.
+        This function should be identical to _write with the exception of an additional 0.03 s in wait time
+        """
+        cmds = self._format_string(base_str, **kwargs)
+        for cmd in cmds.split(";"):
+            self.instrument.write(cmd)
+            time.sleep(0.02 + len(cmd) / 6000 + 0.03)
+        self._is_error()
+
```
Jasper-Harvey0, line 219: Because this is basically a copy paste of the original write function, why not just add an optional keyword argument to the write function to add an optional extra delay?

```python
    def _write(self, data, delay=0.0):
        """
        The SPD3303X cannot respond to visa commands as quickly as some other devices
        A 20ms delay was found to be reliable for most commands.
        Note:
        The 6000 number for the sleep is derived from trial and error. The write calls don't seem to block at the rate
        they write. By allowing 166uS delay for each byte of data then the Funcgen doesn't choke on the next call. A
        flat 20ms is added to allow processing time.
        This is especially important for commands that write large amounts of data such as user arbitrary forms.
        # NOTE: SPD programming tips recommends 10-100ms between write commands
        """
        for cmd in data.split(";"):
            self.instrument.write(cmd)
            time.sleep(0.02 + len(cmd) / 6000 + delay)
        self._is_error()
```

You might be able to then use a lambda function to define the extra delay in the api definition?
```python
("save.group1", lambda data: self.write(data, delay=0.5), "*SAV 1"), # Add extra delay to the write function.
```

I think that solution reduces the amount of repeated code we have lying around, although someone might have other opinions on how to do it differently. 
Regardless of the direction we take, I do not think perpetuating that odd variable sleep time is a good thing.

Riley-King: > Because this is basically a copy paste of the original write function, why not just add an optional keyword argument to the write function to add an optional extra delay?

I did try that, however with how the init_api is written, it would discard any additional arguments added to the 3-item tuple.

> You might be able to then use a lambda function to define the extra delay in the api definition?
> 
> ```python
> ("save.group1", lambda data: self.write(data, delay=0.5), "*SAV 1"), # Add extra delay to the write function.
> ```

I did not consider using a lambda to "contain" the `delay` kwarg. If you would like, I can change this code to use a lambda.

clint-lawrence: I've got no strong opinion, but an alternative to lambda `functools.partial(self.write, delay=0.5)`

daniel-montanari: > I've got no strong opinion, but an alternative to lambda `functools.partial(self.write, delay=0.5)`

This is what I would've suggested if I saw this last week.
Lambda would be fine until you need something more structured.

daniel-montanari:
https://pyvisa.readthedocs.io/en/latest/introduction/communication.html#getting-the-instrument-configuration-right

https://pyvisa.readthedocs.io/en/latest/api/resources.html#pyvisa.resources.SerialInstrument.query_binary_values
We use `query_ascii_values` `query` and `write`
The `write` command doesn't use `query_delay` so adding manual sleeps makes sense to stop spamming.
The other commands shouldn't need manual sleeps as they should already default to sleeping for 100ms. If someone has overridden the default then I think it is on them for setting the delay too short.

Riley-King:
Why are the tests failing here? It is testing code that has already passed testing back in September. Have the tests changed between now and then?

It appears that the test environment has not installed the cmd2.ansi module.

Jasper-Harvey0:
> Why are the tests failing here? It is testing code that has already passed testing back in September. Have the tests changed between now and then?
> 
> It appears that the test environment has not installed the cmd2.ansi module.


See changelog for cmd2:

image

https://github.com/python-cmd2/cmd2/blob/main/CHANGELOG.md

PR #229, opened 2025-09-05
Fix cases where the FTDI driver crashes test scripts preventing the power supply from being shut down

Riley-King:
If the FTDI driver raises an exception, it raises a `BaseException` which is a root exception that `catch Exception as e` **WILL NOT** catch. This will kill *every* script if this occurs and the shutdown functions will not be called which has resulted in power supplies being left powered. This does not necessarily fix the power supply being left on but it does give the scripts a fighting chance.

The second commit 948f4d51dcf95ceab9c6ab362bd6dc504adc05c4 takes this a step further and removes **ALL** uses of `BaseException` as it offers no advantage over `Exception` unless we are trying to kill the script.

See [Why is it recommended to derive from Exception instead of BaseException class in Python?](https://stackoverflow.com/a/27995140).

### DIFF
--- a/src/fixate/core/exceptions.py
+++ b/src/fixate/core/exceptions.py
@@ -1,20 +1,20 @@
-class SequenceAbort(BaseException):
+class SequenceAbort(Exception):
     pass
 
 
-class ScriptError(BaseException):
+class ScriptError(Exception):
     pass
 
 
-class TestRetryExceeded(BaseException):
+class TestRetryExceeded(Exception):
     pass
 
 
-class TestAbort(BaseException):
+class TestAbort(Exception):
     pass
 
 
-class InstrumentError(BaseException):
+class InstrumentError(Exception):
     pass
 
 
@@ -22,11 +22,11 @@ class InstrumentTimeOut(InstrumentError):
     pass
 
 
-class ParameterError(BaseException):
+class ParameterError(Exception):
     pass
 
 
-class InvalidScalarQuantityError(BaseException):
+class InvalidScalarQuantityError(Exception):
     pass
 
 
@@ -38,7 +38,7 @@ class InstrumentNotConnected(InstrumentError):
     pass
 
 
-class NotCompatible(BaseException):
+class NotCompatible(Exception):
     pass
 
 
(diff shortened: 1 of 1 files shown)


src/fixate/core/exceptions.py
```diff
@@ -1,32 +1,32 @@
-class SequenceAbort(BaseException):
+class SequenceAbort(Exception):
```
clint-lawrence, line 1: Since these are all getting touched, It's probably worth making a top level `FixateError` and deriving all the more specific errors from that.

src/fixate/core/exceptions.py
```diff
@@ -1,32 +1,32 @@
...
     pass
 
 
-class ScriptError(BaseException):
+class ScriptError(Exception):
     pass
 
 
-class TestRetryExceeded(BaseException):
+class TestRetryExceeded(Exception):
```
clint-lawrence, line 9: Perhaps it grows the scope of this PR too much, but generally python exceptions will be suffixed with `Error`. We could rename them, keeping an alias for compatibility (but probably not worth the churn)

clint-lawrence (commented):
This is a good move in general and it is probably O.K. to just makes this change as is. However the exception handling in the sequencer is also a mess, and _catches_ `BaseException` in a few places. Which is probably worse than raising some base exceptions :(

So I wonder if we should also review the sequencer and check all the except statements also?

Riley-King:
@clint-lawrence I don't think that the sequencer should have the `except BaseException` changed to use `Exception` because if the program were to be killed in some way (e.g; Ctrl+C, `sys.exit`, etc), the application would immediately crash and not run the teardown functions for the test, which could leave a power supply energized. I do agree that catching all exceptions is generally bad practice but I believe it to be justified in this case.

Ideally, Fixate would have a resource manager that uses the `__enter__` and `__exit__` metafunctions - this would mean that even if an exception is not caught, the cleanup function would still run. The hardest part of this would be updating existing scripts to use it, though this could be deferred to an as-needed basis provided that the underlying drivers are still exposed. It would save some of the boilerplate cleanup code used in scripts and make opening and closing resources more consistent.

clint-lawrence:
> I do agree that catching all exceptions is generally bad practice but I believe it to be justified in this case.

Perhaps, but the sequencer is still a mess. And I'm fairly sure catches `BaseException` more than it should. However, I'm working of fuzzy memory, rather than rigorous analysis :)

There is one spot that might be worth checking. In this block, the sequencer catches `BaseException` and re-raises `AbortException`. I'm not sure how that then gets handled further up the call stack. It's possible changing the base class of `AbortException` will change how that works.
https://github.com/PyFixate/Fixate/blob/9f426460e909999a7baa44ce6c44c321f947a709/src/fixate/sequencer.py#L370

jcollins1983:
Because this is in a fork that we don't have access to, this work will continue in #248

PR #232, opened 2025-09-19
Fixing issues in driver tests

Riley-King:
Changed resistance tolerances in keithley 6500 and fluke 8846A tests from 1 ohm to 5 to account for the fact that it is measuring a 100 ohm resistor, fixing it complaining that it got 102 ohms on a two-wire measurement when it expected 100.

Fixed handling of floating-point measurements in keithley 6500 and fluke 8846A tests, changing `min < avg < max` to `min <= avg <= max`.

Marked temperature test in keithley 6500 as xfail. This is to preserve consistency with the fluke 8846A tests as well as eliminate testing a feature that is not used nor supported.

Changed tolerance in Siglent SPD3303X driver voltage test from 10 mV to 50 mV as that is the limit of the devices precision as guaranteed by Siglent anyway.

### DIFF
--- a/test/drivers/test_fluke_8846A.py
+++ b/test/drivers/test_fluke_8846A.py
@@ -9,7 +9,7 @@ load_config()  # Load fixate config file
 # Test values for measurement functions:
 # These are mostly defined either by J413 or an arbitrary number I picked.
 TEST_RESISTANCE = 100  # Resistance in loopback jig for testing
-TEST_RESISTANCE_TOL = 1  # 1 Ohm absolute tolerance
+TEST_RESISTANCE_TOL = 5  # 1 Ohm absolute tolerance
 TEST_CAPACITANCE = 4.7e-6  # Capacitance in loopback jig for testing
 TEST_CAPACITANCE_TOL = 0.5e-6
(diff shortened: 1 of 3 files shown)


test/drivers/test_fluke_8846A.py
```diff
@@ -494,7 +494,7 @@ def test_min_avg_max(mode, samples, nplc, dmm, rm, funcgen):
     avg_val2 = values.avg
     max_val2 = values.max
 
-    assert min_val2 < avg_val2 < max_val2
+    assert min_val2 <= avg_val2 <= max_val2
```
Overlord360: as above

test/drivers/test_keithley_6500.py
```diff
@@ -502,7 +502,7 @@ def test_min_avg_max(mode, samples, nplc, dmm, rm, funcgen):
     avg_val2 = values.avg
     max_val2 = values.max
 
-    assert min_val2 < avg_val2 < max_val2
+    assert min_val2 <= avg_val2 <= max_val2
```
Overlord360: as above

jcollins1983:
Because this is in a fork that we don't have access to, this work will continue in #246

PR #233, opened 2025-09-19
Added python 3.13 to test matrix in Github workflow

Riley-King:


Overlord360:
This was done in 5f0e979
This is no longer required

PR #244, opened 2026-02-24
Mux checks

daniel-montanari:
Remove requirement to have both pin_list and pin_map which had issues in the new switching module if pins were missing from the list.

Add check that pins are unique across muxes.


daniel-montanari:
Having another look at this, I'm not sure whether we want to enforce explicitly including the pin_list or not.
One option is as I have done here where the map_list is used to calculate the pins.
The other option is enforce including pin_list then use that to audit the map_list.
The current behaviour is to calculate the pins list for the mux, then use the supplied pin list for the address handler.

I think it is better to require both since with larger muxes it can be quite easy to stuff up the pins.

PR #245, opened 2026-02-24
Force reset

daniel-montanari:
Jig reset would not do anything if the pins were in an undefined state.
Change reset to always update the pins regardless of assumed state.


src/fixate/_switching.py
```diff
@@ -619,13 +619,13 @@ def _do_pending_updates(self) -> None:
...
 
-    def _dispatch_pin_state(self, new_state: PinSetState) -> None:
+    def _dispatch_pin_state(self, new_state: PinSetState, force: bool = False) -> None:
         # check all pins actually have an address handler to send to
         if unknown_pins := (new_state.on | new_state.off) - self._all_pins:
             raise ValueError(f"Can't switch unknown pin(s) {', '.join(unknown_pins)}.")
 
         new_active_pins = (self._active_pins | new_state.on) - new_state.off
-        if new_active_pins != self._active_pins:
+        if (new_active_pins != self._active_pins) | force:
```
jcollins1983: What's the reasoning behind using the bitwise or `|` instead of boolean `or`? This looks like a boolean construct to me.

jcollins1983 (dismissed):
Just one minor thing.

clint-lawrence:
Adding a test would be nice

src/fixate/_switching.py
daniel-montanari: No idea I've changed to `or`

daniel-montanari:
Old version of `jig.reset` fails the test due to not switching the pins.
New version passes the test.

Jasper-Harvey0 (approved):
Looks good to me. Tested on physical jig too.

PR #251, opened 2026-03-11
Fix file export in waveform values

Jasper-Harvey0:
I forgot a \n on the csv line write.
When exporting the data to a file, you end up with one big long line separated by commas instead of many lines.


jcollins1983 (approved):
Looks good.

PR #252, opened 2026-03-11
DSO Driver - Implement Invert Channel Function

Jasper-Harvey0:
The DSO Driver does not currently have an invert channel function. This change just implements that.

This is the visa command that is sent when you call dm.dso.ch1.invert(True):

image


jcollins1983 (approved):
Makes sense.

Jasper-Harvey0:
This is duplicated in #253 because I can't git very well. Closing issue.

PR #253, opened 2026-03-12
Add new DSO functions

Jasper-Harvey0:
Invert and Vtime functions are not currently implemented in the DSO function. 
This PR is just adding them.


src/fixate/drivers/dso/helper.py
```diff
@@ -704,6 +704,16 @@ def __init__(self):
         self.xmax = MeasureAllSources()
         self.xmin = MeasureAllSources()
 
+    def vtime(self, time: float, channel: int):
```
jcollins1983, line 707: This breaks the methodology used in definin all the other functions, but it's easier to understand since you don't have to go to a million other places to see what's going on, so I'm not opposed to the break in paradigm.
I'm assuming that implementing this is what led you to make the changes in #255

jcollins1983 (approved):
This one seems to duplicate content from PR #252, did you branch from that branch for this one?
At first I thought `vtime` was a horrible name for a function but it looks like you're just mirroring the VISA function name anyway, so I guess it makes sense.
Putting the break in paradigm aside, I think the change look reasonable.

src/fixate/drivers/dso/helper.py
Jasper-Harvey0: Yeah it does a bit... That's mainly because the existing measure.xxx.ch() definition doesn't allow you to enter arguments into the call. I could have solved it by making more classes... That spawned my tangent into the re-write yes. 

I'll leave it for now, and have a think about retaining the current paradigm when I flesh out the changes in #255.

Jasper-Harvey0:
In regard to duplicate changes; I re-named the branch, but it didn't seem to pick up the change on the github side of things. I will close the original PR and use this to bring in both changes.

PR #254, opened 2026-03-12
Fix ret_value checks

Jasper-Harvey0:
Bjorn pointed out that the get float function does not work when given 0.0 or 0 as a value. 
Investigation shows that the return value checks simply ``is ret_val`` which evaluates to false on 0.0 as its a falsy value.

I have updated the implementation in the ui to explicitly check that ret_val is not None instead.


clint-lawrence:
Does this fix this issue also? https://github.com/PyFixate/Fixate/issues/213

jcollins1983:
I believe all of these things are fixed in my branch that I'm going to revive and get finished #214

Jasper-Harvey0:
Awesome, I'll close.

Jasper-Harvey0:
To be completed in https://github.com/PyFixate/Fixate/pull/214

PR #255, opened 2026-03-13
Driver update concept for DSO

Jasper-Harvey0:
This is my idea to revamp the DSO driver. 
The implementation intends to keep the same API so we don't have to change test scripts.

Basically it takes a more direct approach to constructing classes in a fashion that reflects the old API. 
We get type checking, code completion, and (I think) better maintainability. 
This will also allow us to make a bit more of a distinction between oscilloscopes that have different channel counts, by raising errors if an instrument does not have a channel.


clint-lawrence:
Sounds like a super idea!

PR #256, opened 2026-03-17
Add FTDI 232h driver

jcollins1983:
Because of issues with libusb and the use of zadig this driver was created to replace pyftdi. It wraps an FTDI provided driver [libmpsse](https://ftdichip.com/wp-content/uploads/2025/08/libmpsse-windows-1.0.8.zip) that builds on the lower level FTD2XX driver.

FTDI provide the source code along with the DLL, which is good because I modified it to include an open by description function that didn't come in the box. The modified source code could possibly be added as a separate repo under the organisation (I did turn it into a git project before making modifications, so the change history is there.

The structure of this library was fashioned on the pyftdi library, mostly to reduce the impact on jigs that use it.

This has been tested on the jigs that uses the pyftdi library and there is parity in performance between the two for the I2C read_from/write_to functions that are used. No unit tests were added because the tests are hardware in the loop and would require a slave device to get responses from etc


clint-lawrence:
This would make a nice independent package, especially given the api matches pyftdi ;)

My only comment is around the logistics. Are we allowed to redistribute the dll? And how was the original source from ftdi license? bundling the dll into the package isn't great, but I understand it simplifies distribution.

jcollins1983:
> This would make a nice independent package, especially given the api matches pyftdi ;)
> 
> My only comment is around the logistics. Are we allowed to redistribute the dll? And how was the original source from ftdi license? bundling the dll into the package isn't great, but I understand it simplifies distribution.

Yeah, I did have an internal battle about including in Fixate or making a separate package. I'm happy to yank it out into its own project then include it as a dependency in amptest.

The answer on the redistribution of the DLL it is its use in projects is expressly allowed in the source code readme.

From ReadMe.txt in the source code in the zip of the link I provided: 

> The pre-built lib and dll files can be used directly in a project. They are not
signed by FTDI. The source code is included to allow the libraries to be 
rebuilt and signed or integrated in another project.

Also this statement in the source code:
> FTDI DRIVERS MAY BE DISTRIBUTED IN ANY FORM AS LONG AS LICENSE INFORMATION IS NOT MODIFIED.

jcollins1983:
Talking to @Jasper-Harvey0 and @daniel-montanari the general consensus is to move this out into its own package. So, I'll kill this PR and create a new repo in our private enterprise.

clint-lawrence:
> the general consensus is to move this out into its own package. So, I'll kill this PR and create a new repo in our private enterprise.

It could be a public repo under the PyFixate org

PR #257, opened 2026-06-01
Added DC Load driver for Rigol DL3021

ckukas-ampcontrol:
Added DC Load driver for Rigol DL3021


src/fixate/drivers/dcload/rigol_dl3021.py
```diff
@@ -0,0 +1,97 @@
+
+from typing import Any
+from fixate.core.exceptions import InstrumentError, ParameterError
+from fixate.drivers.dcload.helper import DCLoad
+from strenum import StrEnum
```
daniel-montanari: This is a 3rd party library designed for older versions of python.
Can change this to `from enum import StrEnum`

test/drivers/J413.py
```diff
@@ -33,6 +34,13 @@ def dmm(self):
             self._dmm = dmm.open()
             self._dmm.instrument.timeout = 7000
         return self._dmm
+    
+    @property
+    def dcload(self):
+        if self._dcload is None:
+            self._dcload = dcload.open()
+            self._dcload.instrument.timeout = 7000
```
daniel-montanari: Whats the reason for setting the timeout?

src/fixate/drivers/dcload/helper.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def reset(self) -> None:
+        # Reset the instrument to a known state. For example, this might be called in a fixture finalizer to ensure the instrument is left in a safe state.
+        ...
+
+    def get_identity(self) -> str:
+        # Return instrument identity string (for example, the *IDN? response).
+        ...
+
+    def set_mode(self, mode) -> None:
```
daniel-montanari: Should add an annotation for mode

test/conftest.py
daniel-montanari, line 1: Looks good

src/fixate/drivers/dcload/rigol_dl3021.py
daniel-montanari: May also want to move the mode up into the helper file to help with annotations

src/fixate/drivers/dcload/rigol_dl3021.py
```diff
@@ -0,0 +1,97 @@
...
+    REGEX_ID = "DL3021"
+    INSTR_TYPE = "VISA"
+    write_termination = "\n"
+    read_termination = "\n"
+
+    def __init__(self, instrument: Any) -> None:
+        self.instrument = instrument
+        self.instrument.timeout = 1000
+        self.instrument.read_termination = self.read_termination
+        self.instrument.write_termination = self.write_termination
```
daniel-montanari: Possibly can have read/write termination inside the init since we never use them again.
Do we actually have to change this from the default setting?

src/fixate/drivers/dcload/rigol_dl3021.py
```diff
@@ -0,0 +1,97 @@
...
+        return self.write(f":SOUR:CURR:RANG {current_range}")
+    
+    def get_current_range(self) -> float:
+        # Get the current range in Amps
+        return float(self.query(":SOUR:CURR:RANG?").strip())
+    
+    def set_current(self, current: float) -> None:
+        # Set the current to the specified value in Amps 
+        mode = self.get_mode()
+        if mode != 'CC':
```
daniel-montanari: Is the mode different to the mode enum at the top? Or are there more abbreviations?

daniel-montanari (commented):
Generally looks good, need to set up pre-commit.

test/drivers/J413.py
ckukas-ampcontrol: Copy and paste from DMM - have removed so it uses the default setting

src/fixate/drivers/dcload/helper.py
```diff
@@ -0,0 +1,42 @@
...
+    def get_identity(self) -> str:
+        """Return instrument identity string."""
+        ...
+
+    def set_mode(self, mode: Mode) -> None:
+        """Set the mode of the load (for example, constant current, constant voltage, etc.)."""
+        ...
+
+    def set_enabled(self, enable: bool) -> None:
+        """Enable (1) or disable (0) the load."""
```
daniel-montanari: If this is a bool it should be True/False

src/fixate/drivers/dcload/helper.py
```diff
@@ -0,0 +1,42 @@
...
+# List of modes that an electronic load can be set to
+Mode = Literal[
+    "CONSTANT_CURRENT",
+    "CONSTANT_VOLTAGE",
+    "CONSTANT_RESISTANCE",
+    "CONSTANT_POWER",
+]
+
+
+class DCLoad(Protocol):
```
daniel-montanari: I think we also need to add this.
`from fixate.drivers import DriverProtocol`
Then update the helper class
`class DCLoad(DriverProtocol, Protocol):`

This will sync up the `log_instrument_open` call which doesn't use the helper.

src/fixate/drivers/dcload/rigol_dl3021.py
```diff
@@ -0,0 +1,101 @@
...
+        elif mode == "CONSTANT_RESISTANCE":
+            scpi_mode = "RES"
+        elif mode == "CONSTANT_POWER":
+            scpi_mode = "POW"
+        else:
+            raise ParameterError(f"Invalid mode: {mode}")
+
+        return self.write(f":SOUR:FUNC {scpi_mode}")
+
+    def get_mode(self) -> str:
```
daniel-montanari: If functions like this are only being used for testing then they should be prefixed with an _ to indicate they are private to this class

src/fixate/drivers/dcload/rigol_dl3021.py
```diff
@@ -0,0 +1,101 @@
...
+class RigolDL3021(DCLoad):
+    """Driver for Rigol DL3021 DC electronic load."""
+
+    REGEX_ID = "DL3021"
+    INSTR_TYPE = "VISA"
+
+    def __init__(self, instrument: Any) -> None:
+        self.instrument = instrument
+        self.instrument.timeout = 1000
+        self._set_current_range = None
```
daniel-montanari: `self._set_current_range:float|None = None`

daniel-montanari (commented):
Can you bump the version to 0.6.5 in src/fixate/__init__.py
Update docs/release-notes.rst

docs/release-notes.rst
```diff
@@ -5,14 +5,15 @@ Release Notes
 *************
 Version 0.6.5
 *************
-Release Date xx-xx-xxxx
+Release Date 03/06/2026
```
daniel-montanari: roll this back, we still don't have a plan to release this yet.

src/fixate/drivers/dcload/rigol_dl3021.py
```diff
@@ -30,44 +30,44 @@ def query(self, command: str) -> str:
...
+        state = self._query(":SOUR:INP:STAT?").strip()
         return state == "1"
 
-    def get_current(self) -> float:
+    def _get_current(self) -> float:
         """Get the current value in Amps."""
-        return float(self.query(":SOUR:CURR:LEV:IMM?").strip())
+        return float(self._query(":SOUR:CURR:LEV:IMM?").strip())
 
     def set_current_range(self, current_range: float) -> None:
```
daniel-montanari: Can this be any value? Or is this like a multimeter where you will only have fixed increments you can set this to?

test/drivers/test_rigol_dl3021.py
```diff
@@ -42,7 +43,14 @@ def test_enable_load(dcload: RigolDL3021):
         ("constant_power", "CP"),
     ],
 )
-def test_set_mode(dcload: RigolDL3021, mode, expected):
+def test_set_mode(
+    dcload: RigolDL3021,
+    mode: Literal["constant_current"]
```
daniel-montanari, line 48: Just an fyi you can do `Literal["a", "b"]` instead of `Literal["a"] | Literal["b"]`

daniel-montanari (approved):
Looks good.

daniel-montanari:
Blocked by https://github.com/PyFixate/Fixate/pull/259

src/fixate/drivers/dcload/rigol_dl3021.py
daniel-montanari: Current level depends on the instrument used 4/40 Amps or 6/60 Amps.

PR #261, opened 2026-06-18
Upgrade formatters

daniel-montanari:
Black was on V22 which can't parse newer python syntax.
Sync formatting with pre-commit.


daniel-montanari:
I'm seeing a lot of deprecation warnings coming from pre-commit actions. I think its probably simpler to just manually bump both pre-commit and tox to use the same newer version of black.

daniel-montanari:
So I did this without seeing https://github.com/PyFixate/Fixate/pull/214#discussion_r3293861338

Can delete this if you like. It might be a pain to merge in and not worth it.

jcollins1983:
I definitely don't want to keep black in tox. If we're going to have anything in tox, I would prefer it be a pre-commit call so that there is only one place to update the version (.pre-commit-config.yaml) . As I mentioned in [this comment](https://github.com/PyFixate/Fixate/pull/214#issuecomment-4737133097) I didn't do this because I would personally prefer to defer changes to code to when there are actual modifications being done, but also happy include calls to pre-commit in tox and/or the workflows and just wear the extra overhead in reviewing the formatting changes that are extraneous to the real change being reviewed at the time, that we the entire codebase remains compliant to the version of the formatter currently in use.

tox.ini
jcollins1983: You've left the black environment there.

jcollins1983 (changes requested):
It looks ok, though I don't think we need to use an action for pre-commit. If we use a tox environment, we will be able to run pre-commit alongside all other tests just by running `tox`. Though anyone making changes should really have the pre-commit hook installed anyway.
And it means that everything is still contained in the tox.ini as far as what is running without having to refer to the CI setup.

If you want to push ahead with the changes as they are, can you please delete the reference to black in the envlist in tox.ini file?

tox.ini
daniel-montanari: Yeah just realised, removed.

jcollins1983 (approved):
Looks good

tox.ini
jcollins1983: Ok

PR #265, opened 2026-07-21
UI changes

Jasper-Harvey0:
- Updated font size on the user interaction buttons
- Sequence end now shows "TEST PASS" or "TEST FAIL" in the image scene to make things really obvious.

image

image

### DIFF
--- a/src/fixate/ui_gui_qt/ui_gui_qt.py
+++ b/src/fixate/ui_gui_qt/ui_gui_qt.py
@@ -266,7 +266,6 @@ class FixateGUI(QtWidgets.QMainWindow, layout.Ui_FixateUI):
         :param path: Relative path to image within the test scripts package
         :return: None
         """
-        self.ImageView.setBackgroundBrush(QtGui.QBrush())
         try:
             image_data = pkgutil.get_data("module.loaded_tests", path)
         except (FileNotFoundError, OSError):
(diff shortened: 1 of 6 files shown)


Jasper-Harvey0:
We should have updated the version to 0.7.0 because of the python update....

Jasper-Harvey0:
Oops, I have added the changes for adding user-id flag... @jcollins1983 I can either prune and force push, or we can do both changes in this one PR?

jcollins1983:
> Oops, I have added the changes for adding user-id flag... @jcollins1983 I can either prune and force push, or we can do both changes in this one PR?

It's not worth the churn to split it out. I'll review it under this one.

src/fixate/ui_gui_qt/ui_gui_qt.py
```diff
@@ -353,6 +355,29 @@ def on_image_clear(self):
         self.ImageView.set_scene(self.image_scene)
         self.ImageView.setScene(self.image_scene)
 
+    def _show_sequence_result(self, text, fill_colour):
+        self.on_image_clear()  # Clear the scene
```
jcollins1983: These 2 comments don't really add any value. A brief docstring at the top that describes what the whole method is doing would be more valuable IMO.

jcollins1983 (dismissed):
Looks ok. Just one comment around the commenting in the new method in the UI if you care to take it on board.

Jasper-Harvey0:
Comments have been updated. Note that the version is going to 0.7.0 because we should have done that last time...

jcollins1983 (approved):
Looks good.

PR #266, opened 2026-07-29
change test pass and test fail colors to match the colours used elsewhere

Overlord360:
Use the same green and red we use in the test list for the test pass / fail


Overlord360:
Technically Qt.green is just #00FF00 and Qt.red is just #FF0000 but might as well dynamically get the values in case that changes in future. Therefore, it'll match the test list colours even if the Qt constants were to suddenly change for some reason.

